PR #2, opened 2025-02-05
fix: uname is more common than hostname

t-h2o:
- uname is a GNU coreutils
- hostname is a GNU inetutils

coreutils is more common than inetutils

an explanation of the difference on this forum [1]

[1]: https://askubuntu.com/a/294584

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Set build environment variables
         run: |
           echo GOVERSION=$(go env GOVERSION) >> $GITHUB_ENV
-          echo BUILD_HOST=$(hostname) >> $GITHUB_ENV
+          echo BUILD_HOST=$(uname -n) >> $GITHUB_ENV
           echo BUILD_USER=$(whoami) >> $GITHUB_ENV
       - name: Release with goreleaser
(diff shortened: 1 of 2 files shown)


alexlovelltroy:
The reference indicates that on Linux, these are the same.  Is there an issue you're seeing that indicates `hostname` is problematic?

t-h2o:
Just in case that the `inetutils` package is not installed on the machine, `hostname` will not be found.

davidallendj:
Has this been a problem for you in the past? Would it be possible to just install `inetutils`?

synackd:
While I can see the practicality of using `uname -n` over `hostname` within the Makefile, which runs on user systems which may or may not have `inetutils` installed, I don't see the practicality of replacing it in GitHub Actions, which always runs with `ubuntu-latest`, which _does_ have it installed. Not that it is incorrect, but it doesn't make sense to me to replace it in GitHub Actions and not in the Makefile. In fact, _practically_, it would make sense to do the opposite.

I am not against changing `hostname` to `uname -n`, but if this is to be done, then we should change it everywhere, _especially_ in the Makefile.

alexlovelltroy:
I'm inclined to close this PR.  It seems like a reasonable idea in some cases, but isn't needed for our github actions.

PR #3, opened 2025-02-05
Add optional magellan discovery commands

davidallendj:
This PR adds the ability to compile the `magellan` commands into the `ochami` executable via a `--tags` flag. To add the `magellan` commands to the build, run `go build` with `--tags=magellan`. The commands can be found with `ochami discover` and are used similarly to the standalone binary.

For example,

```bash
ochami discover scan \
    --subnet 172.16.0.0 \
    --subnet-mask 255.255.255.0 \
    --format json \
    --cache data/assets.db \
```
...is equivalent to the following...
```bash
./magellan scan \
    --subnet 172.16.0.0 \
    --subnet-mask 255.255.255.0 \
    --format json \
    --cache data/assets.db \
```
Any fixes and updates applied to the upstream `magellan` with subsequently be applied to the `ochami discover` equivalent as well without any additional modifications needed.

### DIFF
--- a/cmd/magellan.go
+++ b/cmd/magellan.go
@@ -1,0 +1,15 @@
+//go:build magellan || all
+// +build magellan all
+
+package cmd
+
+import (
+	magellan_cmd "github.com/OpenCHAMI/magellan/cmd"
+)
+
+func init() {
+	discoverCmd.AddCommand(magellan_cmd.ScanCmd)
+	discoverCmd.AddCommand(magellan_cmd.CollectCmd)
+	discoverCmd.AddCommand(magellan_cmd.ListCmd)
+	discoverCmd.AddCommand(magellan_cmd.CrawlCmd)
+}


cmd/magellan.go
```diff
@@ -0,0 +1,15 @@
+//go:build magellan || all
+// +build magellan all
```
synackd, line 2: Right now these are the only tag definitions in `ochami`. We should figure out a tag structure that makes sense for having Magellan be optional. Do we want it to be built-in by default or should the builder specifically specify it?

synackd (commented):
Besides my comment above, just a couple more things:

- We should run `go mod tidy` to include the Magellan modules.
- CGO presents a non-trivial concern.

  Magellan uses it for sqlite and `ochami` does not use it at all. Building `ochami` without CGO and with magellan will cause `ochami discover scan` to fail:

  ```
  [user@box ochami]$ CGO_ENABLED=0 make GOFLAGS='-tags=magellan'
  [user@box ochami]$ ./ochami discover scan --subnet 172.16.0.0 --subnet-mask 255.255.255.0
  DEBUG | root.go:116 > logging has been initialized
  panic: Binary was compiled with 'CGO_ENABLED=0', go-sqlite3 requires cgo to work. This is a stub
  
  goroutine 1 [running]:
  github.com/jmoiron/sqlx.MustExec(...)
      /home/user/go/pkg/mod/github.com/jmoiron/sqlx@v1.4.0/sqlx.go:725
  github.com/jmoiron/sqlx.(*DB).MustExec(0xd138f8?, {0xc0008680d0?, 0xc00002da00?}, {0x0?, 0x1?, 0xc000418038?})
      /home/user/go/pkg/mod/github.com/jmoiron/sqlx@v1.4.0/sqlx.go:369 +0x5c
  github.com/OpenCHAMI/magellan/internal/cache/sqlite.CreateScannedAssetIfNotExists({0xc00002da00, 0x1d})
      /home/user/go/pkg/mod/github.com/!open!c!h!a!m!i/magellan@v0.1.8/internal/cache/sqlite/sqlite.go:30 +0xe5
  github.com/OpenCHAMI/magellan/internal/cache/sqlite.InsertScannedAssets({0xc00002da00?, 0x13?}, {0xc00039a008, 0x6, 0xb3bf60?})
      /home/user/go/pkg/mod/github.com/!open!c!h!a!m!i/magellan@v0.1.8/internal/cache/sqlite/sqlite.go:40 +0x3c
  github.com/OpenCHAMI/magellan/cmd.init.func6(0xc00003e900?, {0xc00036a140?, 0x0?, 0x4?})
      /home/user/go/pkg/mod/github.com/!open!c!h!a!m!i/magellan@v0.1.8/cmd/scan.go:161 +0x832
  github.com/spf13/cobra.(*Command).execute(0x13ad6e0, {0xc00036a100, 0x4, 0x4})
      /home/user/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:989 +0xab1
  github.com/spf13/cobra.(*Command).ExecuteC(0x13a6c80)
      /home/user/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1117 +0x3ff
  github.com/spf13/cobra.(*Command).Execute(...)
      /home/user/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1041
  github.com/OpenCHAMI/ochami/cmd.Execute()
      /opt/shared/ochami/cmd/root.go:64 +0x1a
  main.main()
      /opt/shared/ochami/main.go:8 +0xf
  ```

  Changing the backend storage mechanism for Magellan might be the easiest route, and might be easier than dealing with the cross-compilation complexity of CGO. However, it's worth discussing which is better to pursue.

cmd/magellan.go
davidallendj: The tags are for everything else that's optional and not everything that's meant to be built in by default. So if the PCS was added as optional, the `all` tag would only cover `magellan` and `pcs`. Otherwise, everything that is being built currently without the tags should be the default.

If I want just the default `ochami` commands:
```bash
go build # ...with other options
```

If I want to include *just* the `magellan` commands with default commands:

```bash
go build --tags=magellan
```

If I want to include the hypothetical `pcs` and `magellan` commands:

```bash
go build --tags=all
# ...or...
go build --tags=pcs,magellan
```

Note that the `all` tag would also include any other future commands added as well which should save us from have to include them manually if we intend to build the "full" package.

PR #4, opened 2025-02-14
Add pcs command

cjh1:
This is the start of a pcs command, currently only reports status, but will be built on provide support for power transitions etc.

### DIFF
--- a/cmd/pcs-status.go
+++ b/cmd/pcs-status.go
@@ -1,0 +1,207 @@
+// This source code is licensed under the license found in the LICENSE file at
+// the root directory of this source tree.
+package cmd
+
+import (
+	"errors"
+	"fmt"
+	"github.com/OpenCHAMI/ochami/internal/log"
+	"github.com/OpenCHAMI/ochami/pkg/client"
+	"github.com/OpenCHAMI/ochami/pkg/client/pcs"
+	"github.com/spf13/cobra"
+	"net/http"
+	"strings"
+	"encoding/json"
+	"gopkg.in/yaml.v3"
+	"github.com/elliotchance/pie/v2"
+)
+
+// For now use this to map API name to names that make more sense for the CLI, in
+// the end we might just move these aliases to the service. Note: We don't report
+// status for DistLocking (as the only implementation uses ETCD, so the status
+// is just duplicated) or the TaskRunner (as we only use the local implementation)
+type commandOutput struct {
+	Status         string `json:"pcs,omitempty"`
+	KvStore        string `json:"storage,omitempty"`
+	StateManager   string `json:"smd,omitempty"`
+	Vault          string `json:"vault,omitempty"`
+}
+
+// format commandOutput as JSON or YAML
+func formatOutput(output commandOutput, format string) ([]byte, error) {
+	switch strings.ToLower(format) {
+	case "json":
+		if bytes, err := json.Marshal(output); err != nil {
+			return nil, fmt.Errorf("failed to marshal output into JSON: %w", err)
+		} else {
+			return bytes, nil
+		}
+	case "yaml":
+		if bytes, err := yaml.Marshal(output); err != nil {
+			return nil, fmt.Errorf("failed to marshal output into YAML: %w", err)
+		} else {
+			return bytes, nil
+		}
+	default:
+		return nil, fmt.Errorf("unknown output format: %s", format)
+	}
+}
+
+// Get the status of PCS either "live" or "ready"
+func getStatus(pcsClient *pcs.PCSClient) (string, error) {
+	httpEnv, err := pcsClient.GetReadiness()
+	if err != nil {
+		if errors.Is(err, client.UnsuccessfulHTTPError) {
+			log.Logger.Fatal().Err(err).Msg("PCS status (readiness) request yielded unsuccessful HTTP response")
+		} else {
+			log.Logger.Fatal().Err(err).Msg("failed to get PCS status (readiness)")
+		}
+	}
+
+	// We are in the "ready" state
+	if httpEnv.StatusCode == http.StatusNoContent {
+		return  "ready", nil
+	}
+
+	// If we are not "ready" then check our "liveness"
+	httpEnv, err = pcsClient.GetLiveness()
+	if err != nil {
+		if errors.Is(err, client.UnsuccessfulHTTPError) {
+			log.Logger.Fatal().Err(err).Msg("PCS status (liveness) request yielded unsuccessful HTTP response")
+		} else {
+			log.Logger.Fatal().Err(err).Msg("failed to get PCS status (liveness)")
+		}
+	}
+
+	// We are in the "live" status
+	if httpEnv.StatusCode == http.StatusNoContent {
+		return "live", nil
+	} else {
+		return "", errors.New("unable to get PCS state")
+	}
+}
+
+// struct used to unmarshall /health endpoint response
+type healthOutput struct {
+	KvStore        string
+	DistLocking    string
+	StateManager   string
+	Vault          string
+	TaskRunner     string
+}
+
+// allowed flag for status command
+func flags() []string {
+	return []string{"all", "storage", "smd", "vault"}
+}
+
+// pcsStatusCmd represents the pcs-status command
+var pcsStatusCmd = &cobra.Command{
+	Use:   "status",
+	Args:  cobra.NoArgs,
+	Short: "Get status of PCS service",
+	Run: func(cmd *cobra.Command, args []string) {
+		// Without a base URI, we cannot do anything
+		pcsBaseURI, err := getBaseURI(cmd)
+		if err != nil {
+			log.Logger.Fatal().Err(err).Msg("failed to get base URI for PCS")
+		}
+
+		// Create client to make request to PCS
+		pcsClient, err := pcs.NewClient(pcsBaseURI, insecure)
+		if err != nil {
+			log.Logger.Fatal().Err(err).Msg("error creating new PCS client")
+		}
+
+		// Check if a CA certificate was passed and load it into client if valid
+		useCACert(pcsClient.OchamiClient)
+
+		// Figure out if we need to hit the /health endpoint (only if a flag has been provided)
+		flagsProvided := false
+		flags := flags()
+		for i := 0; i < len(flags); i++ {
+			flagsProvided = flagsProvided || cmd.Flag(flags[i]).Changed
+		}
+
+		var health healthOutput
+		if flagsProvided {
+			healthHttpEnv, err := pcsClient.GetHealth()
+			if err != nil {
+				if errors.Is(err, client.UnsuccessfulHTTPError) {
+					log.Logger.Fatal().Err(err).Msg("PCS status (health) request yielded unsuccessful HTTP response")
+				} else {
+					log.Logger.Fatal().Err(err).Msg("failed to get PCS status (health)")
+				}
+			}
+
+			// Unmarshall the health
+			err = json.Unmarshal(healthHttpEnv.Body, &health)
+			if err != nil {
+				log.Logger.Fatal().Msg("failed to unmarshal health")
+			}
+		}
+
+		var output commandOutput
+		reportPCSState := !flagsProvided
+
+		// Process the flags and copy the parts we need from the /health
+		// endpoint response
+		if cmd.Flag("all").Changed {
+			output = commandOutput{
+				KvStore: health.KvStore,
+				StateManager: health.StateManager,
+				Vault: health.Vault,
+			}
+			reportPCSState = true
+		}
+		if cmd.Flag("storage").Changed {
+			output.KvStore = health.KvStore
+		}
+		if cmd.Flag("smd").Changed {
+			output.StateManager = health.StateManager
+		}
+		if cmd.Flag("vault").Changed {
+			output.Vault = health.Vault
+		}
+
+		// Now deal with the PCS status
+		if reportPCSState {
+			pcsStatus, err := getStatus(pcsClient)
+			if err != nil {
+				log.Logger.Fatal().Err(err).Msg("failed to get PCS status")
+			}
+
+			output.Status = pcsStatus
+		}
+
+		// Print output
+		outFmt, err := cmd.Flags().GetString("output-format")
+		if err != nil {
+			log.Logger.Fatal().Err(err).Msg("failed to get value for --output-format")
+		}
+		if outBytes, err := formatOutput(output, outFmt); err != nil {
+			log.Logger.Fatal().Err(err).Msg("failed to format output")
+		} else {
+			fmt.Println(string(outBytes))
+		}
+	},
+}
+
+func init() {
+	pcsStatusCmd.Flags().Bool("all", false, "print all status data from PCS")
+	pcsStatusCmd.Flags().Bool("storage", false, "print status of storage backend from PCS")
+	pcsStatusCmd.Flags().Bool("smd", false, "print status of PCS connection to SMD")
+	pcsStatusCmd.Flags().Bool("vault", false, "print status of PCS connection to Vault")
+
+	// Mark "all" as mutally exusive of all the other flags
+	// First we need a list of flags without "all"
+	flags := pie.FilterNot(flags(), func(flag string) bool {
+        return flag == "all"
+    })
+	for i := 0; i < len(flags); i++ {
+		pcsStatusCmd.MarkFlagsMutuallyExclusive("all", flags[i])
+	}
+
+	pcsStatusCmd.Flags().StringP("output-format", "F", defaultOutputFormat, "format of output printed to standard output")
+	pcsCmd.AddCommand(pcsStatusCmd)
+}
--- a/cmd/pcs.go
+++ b/cmd/pcs.go
@@ -1,0 +1,31 @@
+// This source code is licensed under the license found in the LICENSE file at
+// the root directory of this source tree.
+package cmd
+
+import (
+	"os"
+
+	"github.com/OpenCHAMI/ochami/internal/log"
+	"github.com/spf13/cobra"
+)
+
+// pcsCmd represents the pcs command
+var pcsCmd = &cobra.Command{
+	Use:   "pcs",
+	Args:  cobra.NoArgs,
+	Short: "Interact with the Power Control Service (PCS)",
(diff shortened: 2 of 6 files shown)


synackd:
This looks excellent. Is this ready for review? Does it need further testing? Also, are you comfortable and willing to create a manual page for `pcs` at `man/ochami-pcs.1.sc` in the [scdoc format](https://man.archlinux.org/man/scdoc.5.en)? An easy way to get started is to use the [**ochami-bss(1)** manual page](https://github.com/OpenCHAMI/ochami/blob/main/man/ochami-bss.1.sc) as a starting point.

davidallendj:
I tried the following command:
```bash
ochami pcs status --base-uri https://$cluster_uri/hsm/v2
```
And got this as output:
```bash
DEBUG | root.go:116 > logging has been initialized
DEBUG | root.go:296 > using base URI passed on command line
DEBUG | root.go:297 > base URI: https://instructors.si.usrc:8443/hsm/v2
DEBUG | client.go:243 > GET: https://instructors.si.usrc:8443/hsm/v2/health
DEBUG | client.go:264 > Request headers:
DEBUG | client.go:266 >   User-Agent: [ochami/unknown]
DEBUG | client.go:275 > No body in request
DEBUG | client.go:286 > Response status: 200 OK
DEBUG | client.go:288 > Response headers:
DEBUG | client.go:290 >   Date: [Wed, 19 Feb 2025 18:15:19 GMT]
DEBUG | client.go:290 >   Content-Length: [0]
DEBUG | client.go:307 > No body in response
INFO  | http.go:143 > Response status: HTTP/2.0 200 OK
FATAL | pcs-status.go:140 > failed to unmarshal health
```

Also as a side note (probably not for this PR), it wasn't initially clear what to use for the `--base-uri` argument. It states using `ochami help`...
```bash
Flags:
  -u, --base-uri string     base URI for OpenCHAMI services
```
So, I'd assume you'd pass a URI like `https://$cluster_uri` as opposed to the root to SMD which is `https://$cluster_uri/hsm/v2`.

synackd:
> Also as a side note (probably not for this PR), it wasn't initially clear what to use for the `--base-uri` argument. It states using `ochami help`...
> 
> ```shell
> Flags:
>   -u, --base-uri string     base URI for OpenCHAMI services
> ```
> 
> So, I'd assume you'd pass a URI like `https://$cluster_uri` as opposed to the root to SMD which is `https://$cluster_uri/hsm/v2`.

That's correct. The service base endpoints (e.g. `/hsm/v2`, `/boot/v1`) are appended to the value of `--base-uri` when making requests to the relevant service.

davidallendj:
> That's correct. The service base endpoints (e.g. `/hsm/v2`, `/boot/v1`) are appended to the value of `--base-uri` when making requests to the relevant service.

Is that why I have to include `/hsm/v2` when using the `pcs` command? I'm guessing it's not being appended when invoking.

synackd:
Right now the base paths for the services are hard-coded in `ochami` (`/hsm/v2` for SMD and `/boot/v1` for BSS). These services not only use those endpoints when running behind an API gateway, but also when running on their own. However, this could change with different API gateway configurations so it might be useful to be able to customize the `ochami` cluster config based on this configuration.

For PCS to match this and have a similar API gateway config to the SMD and BSS, it would need to have its own base path in the service itself, e.g. `/pcs`. Then the `ochami` tool, if following the current paradigm, would set [`basePathPCS`](https://github.com/cjh1/ochami/blob/7b50371ba409e9d13ba05459254209ca61aede7c/pkg/client/pcs/pcs.go#L13) to `/pcs`.

If we wanted to be able to allow the customization of these base paths (e.g. allow custom base paths in the API gateway configuration), then `ochami` should have config options for each service. My inclination is to have defaults, e.g:

| **Service** | **Default Base Endpoint** |
| ----- | ----- |
| SMD | /hsm/v2 |
| BSS | /boot/v1 |
| cloud-init | /cloud-init |
| PCS | /pcs |

and have those be the default if not configured otherwise in the `ochami` config file.

davidallendj:
> > > That's correct. The service base endpoints (e.g. `/hsm/v2`, `/boot/v1`) are appended to the value of `--base-uri` when making requests to the relevant service.
> > 
> > 
> > Is that why I have to include `/hsm/v2` when using the `pcs` command? I'm guessing it's not being appended when invoking. In other words, there should be some logic somewhere that adds it so I don't have to specify it.
> 
> Why do you need `/hsm/v2`, SMD's base path? It looks like there is no base path for this service in this PR.

Whenever I tried without it, I would get a connection refused error.
```bash
[allend@te-head ochami]$ ./ochami pcs status --base-uri https://instructors.si.usrc/
DEBUG | root.go:116 > logging has been initialized
DEBUG | root.go:296 > using base URI passed on command line
DEBUG | root.go:297 > base URI: https://instructors.si.usrc/
DEBUG | client.go:243 > GET: https://instructors.si.usrc/readiness
DEBUG | client.go:264 > Request headers:
DEBUG | client.go:266 >   User-Agent: [ochami/unknown]
DEBUG | client.go:275 > No body in request
FATAL | pcs-status.go:57 > failed to get PCS status (readiness) error="GetReadiness(): error getting PCS liveness: error making GET request to PCS: failed to execute HTTP request: Get \"https://instructors.si.usrc/readiness\": dial tcp 172.16.0.254:443: connect: connection refused"
```

cjh1:
I can definitely add a `/pcs` prefix for pcs. However, to me that seems like a deployment time decision/configuration? If you provide the ability to configure the prefix for each service endpoint, how is that different from just having the `--base-uri` be specific to the service you are interacting with? Or are there commands that interact with multiple services? Another thing to bare in mine is that having a shared `--base-uri` force services under a shared prefix, an alternative deployment might use different domains for each serve?

synackd:
@cjh1 
> I guess I miss understood what `--base-uri` was supposed to be. I assumed that it what the base url for service you are interacting with.

Base *URI* is the URI without any service paths (e.g. `https://foobar.openchami.cluster/`) that the service paths to get appended to. I have a branch I'm working for implementing custom service paths.

> It seem a little brittle to have the prefixes for each service encoded in cli given that this will be very dependent on the deployment. For example the current PCS docker-compose setup doesn't had any prefix.

I agree that having hard-coded paths is brittle and is something I shortcut to get working quickly. As for having service prefixes at all, I'm curious as to how you're currently running PCS with the other OpenCHAMI services. Are you using an API gateway? If so, how are you routing PCS requests to the PCS service? My thought is that having a service prefix makes that API gateway configuration easier, and gives the service its own namespace for API endpoints. Most of the OpenCHAMI (with the exception of OPALL and configurator) services so far (BSS, SMD, cloud-init) follow this paradigm.

I'm also interested in how the base_uri + service_prefix paradigm works with other setups like in k8s. I haven't tested with this and am open to feedback on it.

synackd:
> I can definitely add a `/pcs` prefix for pcs. However, to me that seems like a deployment time decision/configuration? If you provide the ability to configure the prefix for each service endpoint, how is that different from just having the `--base-uri` be specific to the service you are interacting with? Or are there commands that interact with multiple services? Another thing to bare in mine is that having a shared `--base-uri` force services under a shared prefix, an alternative deployment might use different domains for each serve?

The `base-uri` configuration was designed to be a shared base URI that all services would be under, but could be overridden with the `--base-uri` command line flag. Changing the base path for a service would change the path appended to the base URI for a service (e.g. if instead of having SMD configured to run on `/hsm/v2`, one configured it to run at `/apis/hsm/v2`, or something different). So if you're contacting SMD on the full URL `https://demo.openchami.cluster:8443/hsm/v2/Inventory/RedfishEndpoints`, `https://demo.openchami.cluster:8443` would be the base URI, `/hsm/v2` would be the base path for SMD, and `/Inventory/RedfishEndpoints` would be the SMD endpoint used in the library functions of the CLI. These functions concatenate these when making a request: `base-uri` + service base path + final endpoint.

However, it sounds like supporting different domains per-service might be desired and would offer more flexibility for different deployments. My initial thought on how to implement this might be to keep `base-uri` still as a global option within the cluster config, but have optional per-service base URI configurations (e.g. `base-uri-smd`, `base-uri-pcs`) that can be defined to use a different domain for a specific service. Does that seem intuitive or just more complex?

> I am just working with what was already there ( from the upstream ). The service uses a different port, so the docker-compose doesn't need to use a API gateway. I agree that it seems like PCS is an outlier in terms of not having a prefix, so am not adverse to adding it, just think we needed to explore our options.

That's valid. I'm in agreement that we should keep our options open. This tool is meant to be a convenience, after all. 🙂 So, I'm open to feedback on how users expect to configure the tool to interact with a variety of deployments.

cjh1:
> However, it sounds like supporting different domains per-service might be desired and would offer more flexibility for different deployments. My initial thought on how to implement this might be to keep `base-uri` still as a global option within the cluster config, but have optional per-service base URI configurations (e.g. `base-uri-smd`, `base-uri-pcs`) that can be defined to use a different domain for a specific service. Does that seem intuitive or just more complex?

I don't know how desirable a domain per-service deployment is, but it definitely something that people do, so it probably worth having away todo it. Rather than having to enumerate all the services through options like  `base-uri-smd`, `base-uri-pcs` maybe we could just override `--base-uri` at the subcommand level, so something like `ochami pcs --uri <the uri for pcs> ` ( not sure if cobra can support this ). It might also be possible to combine this approach with custom service paths, so you could do `ochami  pcs --uri /pcs` ( in which case `/pcs` could be append to any base-uri configured ) or `ochami  pcs --uri https://pcs.mycooldomain.com` ( in which case the https://pcs.mycooldomain.com would be used ). We would have to figure out how these would map to configuration options for `ochami config`. Thoughts?

synackd:
I like what you recommended and have come up with something that, I think, seems flexible. Let me know your thoughts.

Essentially, `base-uri` becomes `api-uri` and two new config options are introduced: `<service>-path` and `<service>-uri`. All of these are optional but at least `api-uri` and/or `<service>-uri` (where `<service>` is the service name, e.g. `smd` or `bss`) are required. Each of these configuration options have corresponding CLI flags.

For those that want to use the same base URL for all services `api-uri` can be used. Default service paths (e.g. `/hsm/v2` for SMD) are appended to this and can be overridden by `<service>-path` if specified. This is good for those that want to use an API gateway with either a standard (e.g. OpenCHAMI quickstart) or custom configuration.

For cases where the base URI is not necessarily the same per-service (for example if the services are running bare metal on their own port), individual `<service>-uri`s can be used. Even if `api-uri` and/or `<service>-path` is/are specified for the service, the combination of these will always be overridden by `<service>-uri`.

For CLI flags, `api-uri` is the same for both flag and config option. `<service>-uri` and `<service>-path` would become `--uri` and `--path`, respectively, under the relevant `<service>` subcommand. For example, the equivalent to `smd-uri: <val>` in the config file on the command line would be `ochami smd --uri '<val>' ...`. CLI flags override any corresponding directives in the config file.

A cluster config block could look like:

```yaml
clusters:
  - name: foobar
    cluster:
      # If using an API gateway, specify 'api-uri'. Each service will append their
      # base path to this value, and these can be overridden with '<service>-path'.
      api-uri: 'https://foobar.openchami.cluster:8443'

      # If api-uri is used, service paths can be overridden from the defaults.
      smd-path: '/hsm/v2'
      bss-path: '/boot/v1'
      cloud-init-path: '/cloud-init'
      # If the API gateway is configured to pass `/pcs` to PCS, then configure
      # requests to it to be sent to '/pcs'. This is dependent on the API gateway
      # configuration, and PCS need not implement the '/pcs' base path itself.
      pcs-path: '/pcs'

      # Service-specific entire URIs could also be specified using <service>-uri,
      # overriding api-uri and <service>-path, if specified.
      # 
      # E.g. Accessing services directly on their respective ports on the same
      # domain/IP:
      smd-uri: 'https://foobar.openchami.cluster:27779/hsm/v2'
      bss-uri: 'https://foobar.openchami.cluster:27778/boot/v1'
      cloud-init-uri: 'https://foobar.openchami.cluster:27777/cloud-init'
      # Or, something else entirely:
      pcs-uri: 'https://pcs.openchami.cluster'
```

An example of using only flags to set URIs/paths:

```sh
# Configure-less contacting of SMD at:
# https://foobar.openchami.cluster:8443/hsm/v2
ochami --ignore-config smd --api-uri 'https://foobar.openchami.cluster:8443'
ochami --ignore-config smd --uri 'https://foobar.openchami.cluster:8443/hsm/v2'
# Non-standard SMD path:
ochami --ignore-config smd --api-uri 'https://foobar.openchami.cluster:8443' --path '/smd'
ochami --ignore-config smd --uri 'https://foobar.openchami.cluster:8443/smd'
```

## EXAMPLE: Using API Gateway

```yaml
clusters:
    - name: foobar
      cluster:
        # URI of API gateway. Service paths appended to this unless overridden with
        # <service>-api.
        api-uri: https://foobar.openchami.cluster:8443

        # Override default SMD path of `/hsm/v2`. This will direct SMD requests
        # to https://foobar.openchami.cluster:8443/smd
        smd-path: '/smd'
        # No other overrides specified, so default values used. E.g. BSS requests
        # would go to https://foobar.openchami.cluster:8443/boot/v1

        # You can still even override the whole URI of a service. For example, if
        # cloud-init was listening on localhost:27777 (e.g. testing a new version).
        # Note that this overrides the combination of api-uri and smd-base
        # specified above.
        #
        # Still further, this could be overridden on the command line with:
        #
        #  ochami smd --uri 'http://localhost:27777/hsm/v2' ...
        #
        smd-uri: 'http://localhost:27777/hsm/v2'
```

## EXAMPLE: Bare Metal

```yaml
clusters:
    - name: foobar
      cluster:
        # No api-uri specified so each service must have a URL.
        #
        # Example: SMD, BSS, and cloud-init listening on localhost on their default
        # ports.
        smd-uri: http://localhost:27779/hsm/v2
        bss-uri: http://localhost:27778/boot/v1
        cloud-init-uri: http://localhost:27777/cloud-init
        #
        # Example: PCS is listening on an entirely-different URI.
        pcs-uri: https://pcs.openchami.cluster
```

cjh1:
I think this is looking good, that mapping of the service specific URI to the config file works nicely. However, I wonder if we need `<service>-path` and `<service-uri>` to me its a little confusing. `http://localhost:27777/hsm/v2` is a URI and so is `/hsm/v2` ( all be be it a relative one ). My suggest would be to drop `<service>-path` and allow `<service-uri>` to be absolute or relative ( relative to `api-uri` ). I don't think we loose any flexibility and it reduces the number of option a user has to understand. Thoughts?

synackd:
So if I understand correctly, `<service>-path` could be a full URI (`https://...`) or a path URI (`/hsm/v2`). If the former, that full URI overrides any `api-uri` set and is used for that service. If the latter, it requires `api-uri` to be set and gets appended to it.

So for example:

```yaml
clusters:
  - name: demo
    cluster:
      api-uri: https://demo.openchami.cluster:8443
      smd-uri: /smd
```
would use `https://demo.openchami.cluster:8443/smd` as the SMD base URI, whereas:

```yaml
clusters:
  - name: demo
    cluster:
      api-uri: https://demo.openchami.cluster:8443
      smd-uri: https://smd.awesome.cluster/hsm/v2
```
would use `https://smd.awesome.cluster/hsm/v2` as the SMD base URI.

That way, `<service>-uri` can serve two purposes and eliminate the need for two separate config options.

Does that sound correct and reasonable?

synackd:
> I assume you mean `<service>-uri` here?

Whoops, yes I did.

Cool, I think that idea increases the configuration flexibility to support a variety of deployment options without adding too much complexity. I'll get to work filing a ticket documenting this as well as implementing it. Thanks for your feedback!

Also, the manual page looks great!

Since this PR won't work with the current configuration mechanism, I'm thinking we'll need to wait on taking action on it until these configuration changes are implemented. I'm thinking about introducing these config/pcs changes (as well as some others that are impending) in a v0.1 branch. So we may need to rebase/resubmit, but let's wait to decide.

cjh1:
> > I assume you mean `<service>-uri` here?
> 
> Whoops, yes I did.
> 
> Cool, I think that idea increases the configuration flexibility to support a variety of deployment options without adding too much complexity. I'll get to work filing a ticket documenting this as well as implementing it. Thanks for your feedback!
> 

Excellent, I think those will be some great improvements.

> Also, the manual page looks great!
> 
> Since this PR won't work with the current configuration mechanism, I'm thinking we'll need to wait on taking action on it until these configuration changes are implemented. I'm thinking about introducing these config/pcs changes (as well as some others that are impending) in a v0.1 branch. So we may need to rebase/resubmit, but let's wait to decide.

This PR does work with the current configuration mechanism, you just have to provide the full `base-uri`, for example:

```
xxx@openchami:~/work/source/ochami pcs$ ./ochami --base-uri http://localhost:28007  pcs status --all
{"pcs":"ready","storage":"connected, responsive","smd":"connected, responsive","vault":"connected, responsive"}
```

This is running against the `docker-compose.test.integration.yaml` stack in the power-control repo ( with the follow diff to to  expose the port and not run the tests ):

```
diff --git a/docker-compose.test.integration.yaml b/docker-compose.test.integration.yaml
index f60c4d4..c4b5152 100644
--- a/docker-compose.test.integration.yaml
+++ b/docker-compose.test.integration.yaml
@@ -7,17 +7,17 @@ networks:
   pcs:

 services:
-  integration-tests:
-    build:
-      context: .
-      dockerfile: Dockerfile.integration.tests.Dockerfile
-    environment:
-      - API_URL=http://power-control
-      - API_SERVER_PORT=:28007
-    depends_on:
-      - power-control
-    networks:
-      - pcs
+  # integration-tests:
+  #   build:
+  #     context: .
+  #     dockerfile: Dockerfile.integration.tests.Dockerfile
+  #   environment:
+  #     - API_URL=http://power-control
+  #     - API_SERVER_PORT=:28007
+  #   depends_on:
+  #     - power-control
+  #   networks:
+  #     - pcs
   etcd:
     image: quay.io/coreos/etcd:v3.5.17
     environment:
@@ -47,6 +47,8 @@ services:
       - STORAGE=ETCD
       - ETCD_HOST=etcd
       - ETCD_PORT=2379
+    ports:
+      - "28007:28007"
     depends_on:
       - smd
       - etcd

```

synackd (approved):
Just tested and it works. I think that considering that PCS doesn't have an official deployment recipe in the OpenCHAMI deployment-recipes repo yet, this is good to be merged since the URI issue can be mitigated by setting `--base-uri` for now.

PR #8, opened 2025-03-05
feat: improve `config` commands; fix some minor inconveniences

### LINKED ISSUE #6: [DEV] Make configuration flexible to different deployments
## Short Description

The configuration currently is geared toward querying services under a single base URI, e.g. when running an API gateway. As mentioned in #4, `ochami` should support the ability to communicate with services deployed using a variety of deployment methods. In other words, `ochami` should support using different base URIs or path URIs for each service.

See these comments:

- https://github.com/OpenCHAMI/ochami/pull/4#issuecomment-2686572453
- https://github.com/OpenCHAMI/ochami/pull/4#issuecomment-2688512250

## Long Description

The current cluster configuration mechanism looks like this:

```yaml
clusters:
  - name: foobar
    cluster:
      base-uri: https://foobar.openchami.cluster:8443
```

Service base paths (e.g. `/hsm/v2`) are hard-coded and appended to `base-uri`. This pigeonholes deployment options into a specific case if one desires to use `ochami`. A more flexible configuration posed might look like this:

```yaml
clusters:
  - name: foobar
    cluster:
      # URI of API gateway, if used . Per-service `<service>-uri` options either override this option
      # (if the value is a full URI, e.g. 'https://...') or are appended to this option (if the value is a relative
      # path, e.g. '/hsm/v2'). This allows one to not have to repeat URIs, but allows the flexibility to
      # customize both service base paths and entire service URIs.
      api-uri: https://foobar.openchami.cluster:8443

      # Override default SMD base path of `/hsm/v2`. This uses a relative path, so it gets appended to
      # 'api-uri', so SMD requests will get directed to:
      #
      # https://foobar.openchami.cluster:8443/smd
      #
      smd-uri: '/smd'
      #
      # Alternatively, override the entire SMD service URI. This uses a full URI, so it overrides `api-uri`.
      # So, SMD requests will get directed to:
      #
      # https://my.cool.cluster/hsm/v2
      #
      smd-uri: 'https://my.cool.cluster/hsm/v2'
      #
      # No other overrides specified, so default values used. E.g. BSS requests
      # would go to https://foobar.openchami.cluster:8443/boot/v1
```

The above allows one to optionally use `api-uri`, formerly `base-uri`, as a base URI for OpenCHAMI services, e.g. if most services were hosted under it. Various `<service>-uri` options can be used to either change the service base path from the default (e.g. something different than `hsm/v2` for SMD) or override the `api-uri` altogether and use an entirely different URI for a service.

For example, going entirely the opposite direction from using `api-uri`, if all of the OpenCHAMI services were running bare metal on different ports under the same interface, the config might look like:

```yaml
clusters:
  - name: foobar
    cluster:
      smd-uri: https://localhost:27779/hsm/v2
      bss-uri: https://localhost:27778/boot/v1
      cloud-init-uri: https://localhost:27777/cloud-init
      pcs-uri: https://localhost:28007
```

This mechanism expands the deployment options that `ochami` supports:

- Various API gateway configurations (stock or custom, docker compose or Kubernetes and (ideally) beyond)
- Bare metal (e.g. for testing services)

synackd:
Resolves #6 

The linked issue addresses improving `ochami`'s configuration mechanism to support a variety of deployments. Improvement in the `config` subcommands goes a long with that since it allows one to easily modify their configuration file. This PR makes these changes, as well as improves some additional behavior.

# Usage

See `man/ochami-config.1.sc` for more.

```
ochami config [GLOBALOPTS] cluster delete cluster_name
ochami config [GLOBALOPTS] cluster set [-d] cluster_name key value
ochami config [GLOBALOPTS] cluster show [-f format] [cluster_name] [key]
ochami config [GLOBALOPTS] cluster unset cluster_name key
ochami config [GLOBALOPTS] set key value
ochami config [GLOBALOPTS] show [-f format] [key]
ochami config [GLOBALOPTS] unset key
```

GLOBALOPTS:
- `--config <path>` - Use custom config file path
- `--system` - Use system-wide config (`/etc/ochami/config.yaml`)
- `--user` - Use user config (`~/.config/ochami/config.yaml`)

# Changes

## Features

- Move from flags (e.g. `--api-uri`) to keys (e.g. `cluster.api-uri`) in `config` commands for clusters and global config. (Also update manual pages.)

  This allows the available config options to grow without having to add a flag for each option to the various `config` commands.

- Add `config cluster unset` command for unsetting cluster config keys.
- Add `config cluster show` command for showing cluster configuration, for all/individual clusters and all/individual config keys

## Fixes

- Properly make mutually exclusive the `--config`, `--system`, and `--user` options.

   There was an issue making these mutually exclusive in the root command since `--config` is a persistent root flag while `--system` and `--user` are persistent `config` flags. This fix moves this functionality from the root command to each `config` subcommand. It's a bit more verbose than before, but works a lot better.

- Don't ask to create a missing config file if unnecessary.

   For instance, if showing a config value for a specific config file that doesn't exist, it doesn't make sense to create it. However, if setting a config value in a nonexistent config file, it does make sense to create it. Before, `ochami` would ask for _every_ command.

## Refactorization

- Move usage printing with error handling to separate function.
- Use `configFile` variable that the `--config` flag writes to instead of calling `cmd.Flag("config").Value.String()`

### DIFF
--- a/cmd/smd-component-add.go
+++ b/cmd/smd-component-add.go
@@ -31,11 +31,7 @@ This command sends a POST to SMD. An access token is required.`,
 	Run: func(cmd *cobra.Command, args []string) {
 		// Check that all required args are passed
 		if len(args) == 0 && !cmd.Flag("payload").Changed {
-			err := cmd.Usage()
-			if err != nil {
-				log.Logger.Error().Err(err).Msg("failed to print usage")
-				os.Exit(1)
-			}
+			printUsageHandleError(cmd)
 			os.Exit(0)
 		} else if len(args) > 2 {
 			log.Logger.Error().Msgf("expected 2 arguments (xname, nid) but got %d: %v", len(args), args)
--- a/cmd/smd-component-add.go
+++ b/cmd/smd-component-add.go
@@ -22,7 +22,9 @@ var componentAddCmd = &cobra.Command{
 mutually exclusive with the other flags of this command. If - is
 used as the argument to -f, the data is read from standard input.
 
-This command sends a POST to SMD. An access token is required.`,
+This command sends a POST to SMD. An access token is required.
+
+See ochami-smd(1) for more details.`,
 	Example: `  ochami smd component add x3000c1s7b56n0 56
   ochami smd component add --state Ready --enabled --role Compute --arch X86 x3000c1s7b56n0 56
   ochami smd component add -f payload.json
@@ -45,6 +47,7 @@ This command sends a POST to SMD. An access token is required.`,
 		smdBaseURI, err := getBaseURISMD(cmd)
 		if err != nil {
 			log.Logger.Error().Err(err).Msg("failed to get base URI for SMD")
+			logHelpError(cmd)
 			os.Exit(1)
 		}
 
@@ -56,6 +59,7 @@ This command sends a POST to SMD. An access token is required.`,
 		smdClient, err := smd.NewClient(smdBaseURI, insecure)
 		if err != nil {
 			log.Logger.Error().Err(err).Msg("error creating new SMD client")
+			logHelpError(cmd)
 			os.Exit(1)
 		}
 
@@ -90,6 +94,7 @@ This command sends a POST to SMD. An access token is required.`,
 			} else {
 				log.Logger.Error().Err(err).Msg("failed to add component(s) to SMD")
 			}
+			logHelpError(cmd)
 			os.Exit(1)
 		}
 	},
--- a/cmd/smd-component-add.go
+++ b/cmd/smd-component-add.go
@@ -30,10 +30,6 @@ This command sends a POST to SMD. An access token is required.`,
   echo '<json_data>' | ochami smd component add -f -
   echo '<yaml_data>' | ochami smd component add -f - --payload-format yaml`,
 	PreRunE: func(cmd *cobra.Command, args []string) error {
-		// First and foremost, make sure config is loaded and logging
-		// works.
-		initConfigAndLogging(cmd, true)
-
 		// Check that all required args are passed
 		if len(args) == 0 && !cmd.Flag("payload").Changed {
 			printUsageHandleError(cmd)
--- a/cmd/smd-component-add.go
+++ b/cmd/smd-component-add.go
@@ -3,8 +3,8 @@
 package cmd
 
 import (
-	"fmt"
 	"errors"
+	"fmt"
 	"os"
 
 	"github.com/OpenCHAMI/ochami/internal/log"
--- a/cmd/smd-component-add.go
+++ b/cmd/smd-component-add.go
@@ -3,6 +3,7 @@
 package cmd
 
 import (
+	"fmt"
 	"errors"
 	"os"
 
@@ -28,7 +29,7 @@ This command sends a POST to SMD. An access token is required.`,
   ochami smd component add -f payload.yaml --payload-format yaml
   echo '<json_data>' | ochami smd component add -f -
   echo '<yaml_data>' | ochami smd component add -f - --payload-format yaml`,
-	Run: func(cmd *cobra.Command, args []string) {
+	PreRunE: func(cmd *cobra.Command, args []string) error {
 		// First and foremost, make sure config is loaded and logging
 		// works.
 		initConfigAndLogging(cmd, true)
@@ -37,11 +38,13 @@ This command sends a POST to SMD. An access token is required.`,
 		if len(args) == 0 && !cmd.Flag("payload").Changed {
 			printUsageHandleError(cmd)
 			os.Exit(0)
-		} else if len(args) > 2 {
-			log.Logger.Error().Msgf("expected 2 arguments (xname, nid) but got %d: %v", len(args), args)
-			os.Exit(1)
+		} else if len(args) != 2 {
+			return fmt.Errorf("expected 2 arguments (xname, nid) but got %d: %v", len(args), args)
 		}
 
+		return nil
+	},
+	Run: func(cmd *cobra.Command, args []string) {
 		// Without a base URI, we cannot do anything
 		smdBaseURI, err := getBaseURISMD(cmd)
 		if err != nil {
--- a/cmd/smd-component-add.go
+++ b/cmd/smd-component-add.go
@@ -29,6 +29,10 @@ This command sends a POST to SMD. An access token is required.`,
   echo '<json_data>' | ochami smd component add -f -
   echo '<yaml_data>' | ochami smd component add -f - --payload-format yaml`,
 	Run: func(cmd *cobra.Command, args []string) {
+		// First and foremost, make sure config is loaded and logging
+		// works.
+		initConfigAndLogging(cmd, true)
+
 		// Check that all required args are passed
 		if len(args) == 0 && !cmd.Flag("payload").Changed {
 			printUsageHandleError(cmd)
--- a/cmd/smd-rfe-add.go
+++ b/cmd/smd-rfe-add.go
@@ -24,7 +24,9 @@ unless -f is passed to read from a payload file. Specifying -f also is
 mutually exclusive with the other flags of this command and its arguments.
 If - is used as the argument to -f, the data is read from standard input.
 
-This command sends a POST to SMD. An access token is required.`,
+This command sends a POST to SMD. An access token is required.
+
+See ochami-smd(1) for more details.`,
 	Example: `  ochami smd rfe add x3000c1s7b56 bmc-node56 172.16.0.156 de:ca:fc:0f:fe:ee
   ochami smd rfe add -f payload.json
   ochami smd rfe add -f payload.yaml --payload-format yaml
@@ -46,6 +48,7 @@ This command sends a POST to SMD. An access token is required.`,
 		smdBaseURI, err := getBaseURISMD(cmd)
 		if err != nil {
 			log.Logger.Error().Err(err).Msg("failed to get base URI for SMD")
+			logHelpError(cmd)
 			os.Exit(1)
 		}
 
@@ -57,6 +60,7 @@ This command sends a POST to SMD. An access token is required.`,
 		smdClient, err := smd.NewClient(smdBaseURI, insecure)
 		if err != nil {
 			log.Logger.Error().Err(err).Msg("error creating new SMD client")
+			logHelpError(cmd)
 			os.Exit(1)
 		}
 
@@ -78,24 +82,28 @@ This command sends a POST to SMD. An access token is required.`,
 			if cmd.Flag("domain").Changed {
 				if rfe.Domain, err = cmd.Flags().GetString("domain"); err != nil {
 					log.Logger.Error().Err(err).Msg("unable to fetch domain")
+					logHelpError(cmd)
 					os.Exit(1)
 				}
 			}
 			if cmd.Flag("hostname").Changed {
 				if rfe.Hostname, err = cmd.Flags().GetString("hostname"); err != nil {
 					log.Logger.Error().Err(err).Msg("unable to fetch hostname")
+					logHelpError(cmd)
 					os.Exit(1)
 				}
 			}
 			if cmd.Flag("username").Changed {
 				if rfe.User, err = cmd.Flags().GetString("username"); err != nil {
 					log.Logger.Error().Err(err).Msg("unable to fetch username")
+					logHelpError(cmd)
 					os.Exit(1)
 				}
 			}
 			if cmd.Flag("password").Changed {
 				if rfe.Password, err = cmd.Flags().GetString("password"); err != nil {
 					log.Logger.Error().Err(err).Msg("unable to fetch password")
+					logHelpError(cmd)
 					os.Exit(1)
 				}
 			}
@@ -106,6 +114,7 @@ This command sends a POST to SMD. An access token is required.`,
 		_, errs, err := smdClient.PostRedfishEndpoints(rfes, token)
 		if err != nil {
 			log.Logger.Error().Err(err).Msg("failed to add redfish endpoint in SMD")
+			logHelpError(cmd)
 			os.Exit(1)
 		}
 		// Since smdClient.PostRedfishEndpoints does the addition iteratively, we need to deal with
@@ -122,6 +131,7 @@ This command sends a POST to SMD. An access token is required.`,
 			}
 		}
 		if errorsOccurred {
+			logHelpError(cmd)
 			log.Logger.Warn().Msg("SMD redfish endpoint addition completed with errors")
 			os.Exit(1)
 		}
--- a/cmd/smd-rfe-add.go
+++ b/cmd/smd-rfe-add.go
@@ -31,11 +31,7 @@ This command sends a POST to SMD. An access token is required.`,
 	Run: func(cmd *cobra.Command, args []string) {
 		// Check that all required args are passed
 		if len(args) == 0 && !cmd.Flag("payload").Changed {
-			err := cmd.Usage()
-			if err != nil {
-				log.Logger.Error().Err(err).Msg("failed to print usage")
-				os.Exit(1)
-			}
+			printUsageHandleError(cmd)
 			os.Exit(0)
 		} else if len(args) > 4 {
 			log.Logger.Error().Msgf("expected 4 arguments (xname, name, ip_addr, mac_addr) but got %d: %v", len(args), args)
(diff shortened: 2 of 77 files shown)


synackd:
@cjh1 This is the branch [I mentioned](https://github.com/OpenCHAMI/ochami/pull/7#issuecomment-2695399168) with the `config` command improvements.

cjh1 (commented):
Looks like we still have a few references to `base-uri`:

```
cmd/config-cluster-set.go:26:       base-uri: https://foobar.openchami.cluster
man/ochami.1.sc:57:ochami config --user cluster set --default --base-uri https://foobar.openchami.cluster foobar
man/ochami.1.sc:88:*-u, --base-uri* _uri_
```
While testing this out my natural instinct when trying to set the smd URI was to run the following: 

```
ochami config cluster set foobar cluster.smd.uri /hsm/v2
```

Rather than:

```
ochami config cluster set foobar cluster.smd-uri /hsm/v2
```

It seems more intuitive/consistent? In which case the config would look something like:

```yaml
clusters:
    - cluster:
        api-uri: http://localhost:27779
        - smd:
            uri: /hsm/v2
      name: foobar
default-cluster: foobar
log:
    format: rfc3339
    level: warning
```
( looking at it I almost thing dropping the `api-` prefix would be good ... )

This would also open up an easy way to add other service specific configuration without having to add the `<component>-XXX` prefix? Thoughts?

synackd:
> Looks like we still have a few references to `base-uri`:
> 
> ```
> cmd/config-cluster-set.go:26:       base-uri: https://foobar.openchami.cluster
> man/ochami.1.sc:57:ochami config --user cluster set --default --base-uri https://foobar.openchami.cluster foobar
> man/ochami.1.sc:88:*-u, --base-uri* _uri_
> ```

Argh... Missed a few examples in the documentation. Thanks for catching that, I just updated those.

> While testing this out my natural instinct when trying to set the smd URI was to run the following:
> 
> ```
> ochami config cluster set foobar cluster.smd.uri /hsm/v2
> ```
> 
> Rather than:
> 
> ```
> ochami config cluster set foobar cluster.smd-uri /hsm/v2
> ```
> 
> It seems more intuitive/consistent? In which case the config would look something like:
> 
> ```yaml
> clusters:
>     - cluster:
>         api-uri: http://localhost:27779
>         - smd:
>             uri: /hsm/v2
>       name: foobar
> default-cluster: foobar
> log:
>     format: rfc3339
>     level: warning
> ```
> 
> ( looking at it I almost thing dropping the `api-` prefix would be good ... )
> 
> This would also open up an easy way to add other service specific configuration without having to add the `<component>-XXX` prefix? Thoughts?

I actually did (albeit _very_ quickly) consider that and (again, quickly) dismissed it since each service only had one config option. However, I agree with you as that does actually seem quite reasonable and I like the expandability it offers. I think the config as you put it would work except instead of:

```yaml
clusters:
    - cluster:
        api-uri: http://localhost:27779
        - smd:
            uri: /hsm/v2
      name: foobar
```

it would be (no array):

```yaml
clusters:
    - cluster:
        api-uri: http://localhost:27779
        smd:
          uri: /hsm/v2
      name: foobar
```

I think the `--uri` overrides can stay the same (e.g. `ochami smd --uri <smd_uri> ...`).

cjh1:
> I actually did (albeit _very_ quickly) consider that and (again, quickly) dismissed it since each service only had one config option. However, I agree with you as that does actually seem quite reasonable and I like the expandability it offers. I think the config as you put it would work except instead of:
> 
> ```yaml
> clusters:
>     - cluster:
>         api-uri: http://localhost:27779
>         - smd:
>             uri: /hsm/v2
>       name: foobar
> ```
> 
> it would be (no array):
> 
> ```yaml
> clusters:
>     - cluster:
>         api-uri: http://localhost:27779
>         smd:
>           uri: /hsm/v2
>       name: foobar
> ```
Sorry, yes I slipped in a extra - by mistake!

synackd:
So what was `--api-uri` would conflict with each service command's `--uri` flag if it changed to that name. Would it be preferable to have a top-level `--cluster-uri` and keep each service's `--uri`, or make the top-level flag `--uri` and change each service's flag to `--svc-uri`?

synackd:
I went with using `--cluster-uri` for the global and individual `--uri` overrides. ochami-config.5 man page also updated. I also found a bug in the merge routine for merging user config with system config, so got that sorted out!

cjh1:
@synackd Sorry, I will take a look at this, but I am fighting fires else where at the moment, sorry for the delay.

synackd:
@cjh1 Not a problem. I have plenty to do elsewhere here as well. 😵‍💫 This PR is to make life easier, so get to it when you can.

synackd:
@davidallendj @travisbcotton @alexlovelltroy @njones-lanl Your feedback is welcome as well if any of you have time to try it out.

davidallendj:
I haven't tested yet, but is there a way to manually create a config with all default values? I think I prefer to be able to see everything that's available to be configured and go from there.

synackd:
Right now, there's only `ochami config show --ignore-config`. However, that command only prints the default global values (of which there are only two, the log level and format). The rest are cluster config, which doesn't make sense to have a "default" value for. That said, if you look at the **ochami-config**(5) man page (`man/ochami-config.5.sc`) you can see the format as well as examples.

synackd:
Corrected some straggling `api-uri` and `<service>-uri` references, as well as ran `gofmt`.

cjh1:
Hi @synackd, sorry for the delay. I don't think the use of `--uri` on the root command conflicts with using it on the service commands as well, as long as we explicitly check the value on the root command we are good. I tried this with the following patch and unless I am missing something it seems to operate as intended:

```
diff --git a/cmd/lib.go b/cmd/lib.go
index 083fcbd..2ffcf18 100644
--- a/cmd/lib.go
+++ b/cmd/lib.go
@@ -309,9 +309,9 @@ func getBaseURI(cmd *cobra.Command, serviceName config.ServiceName) (string, err
        }
        // 1. Check flags (--cluster-uri and/or --uri) and override any
        // previously-set values while leaving unspecified ones alone.
-       if cmd.Flag("cluster-uri").Changed || (cmd.Flag("uri") != nil && cmd.Flag("uri").Changed) {
+       if rootCmd.Flag("uri").Changed || (cmd.Flag("uri") != nil && cmd.Flag("uri").Changed) {
                log.Logger.Debug().Msg("using base URI passed on command line")
-               ccc := config.ConfigClusterConfig{URI: cmd.Flag("cluster-uri").Value.String()}
+               ccc := config.ConfigClusterConfig{URI: rootCmd.Flag("uri").Value.String()}
                switch serviceName {
                case config.ServiceBSS:
                        ccc.BSS.URI = cmd.Flag("uri").Value.String()
diff --git a/cmd/root.go b/cmd/root.go
index b303c9a..528fc9d 100644
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -57,7 +57,7 @@ func init() {
        rootCmd.PersistentFlags().StringP("log-format", "L", "", "log format (json,rfc3339,basic)")
        rootCmd.PersistentFlags().StringP("log-level", "l", "", "set verbosity of logs (info,warning,debug)")
        rootCmd.PersistentFlags().StringP("cluster", "C", "", "name of cluster whose config to use for this command")
-       rootCmd.PersistentFlags().StringP("cluster-uri", "u", "", "base URI for OpenCHAMI services, excluding service base path")
+       rootCmd.PersistentFlags().StringP("uri", "u", "", "base URI for OpenCHAMI services, excluding service base path")
        rootCmd.PersistentFlags().StringVar(&cacertPath, "cacert", "", "path to root CA certificate in PEM format")
        rootCmd.PersistentFlags().StringVarP(&token, "token", "t", "", "access token to present for authentication")
        rootCmd.PersistentFlags().BoolVarP(&insecure, "insecure", "k", false, "do not verify TLS certificates")
@@ -65,5 +65,5 @@ func init() {
        rootCmd.PersistentFlags().BoolVarP(&config.EarlyVerbose, "verbose", "v", false, "be verbose before logging is initialized")
 
        // Either use cluster from config file or specify details on CLI
-       rootCmd.MarkFlagsMutuallyExclusive("cluster", "cluster-uri")
+       rootCmd.MarkFlagsMutuallyExclusive("cluster", "uri")

```

synackd:
> I'm assuming the config keys that are available to be set are documented somewhere (`--help` flag/man pages)? It wasn't clear to me when I was testing whether or not the examples with `ochami config cluster set` were the only things that could be set. If I do something ridiculous like...
> 
> ```shell
> [allend@re-head ochami]$ ./ochami config --user cluster set allend hello world
> DEBUG | lib.go:101 > logging has been initialized
> DEBUG | config.go:798 > reading config file: /home/allend/.config/ochami/config.yaml
> ERROR | config-cluster-set.go:87 > failed to modify config file error="failed to modify config for existing cluster allend: decoding failed due to the following error(s):\n\n'' has invalid keys: hello"
> ```
> 
> I see the error and expect it to point me in the right direction. Maybe this could be improved to say something like "try `ochami config cluster --help`" or "see specific man page for more info" so we'd know where to look.

I think the help output of `ochami config` commands is the right place to put references to the `ochami-config(5)` manual page. I suppose it wouldn't hurt to put a "see command --help" message after the error message. Cobra commands have a `CommandPath()` method that returns the entire command invocation without flags/args, which could be used to create a generic help-printing log function.

> I know this isn't a part of this PR, but I wasn't expecting the config merging between `/etc/ochami/config.yaml` and `$HOME/.config/ochami/config.yaml`. To me, it doesn't seem like an obvious thing that it's happening and it can catch you by surprise if you skip reading the README.

This could be documented better in the `ochami-config(1)` manual page. I don't think it's out of scope for this PR.

> Also, how do I view the man pages? Do I have to compile them using goreleaser to view them? I didn't see instructions in the README.

The README hasn't been updated in a while. There is [a PR](https://github.com/OpenCHAMI/ochami/pull/9) open to add manual page references to the README as well as build instructions. Building the manual pages uses `scdoc`, and, if you didn't want to use Goreleaser to build an entire release, there is a Makefile target for it:

```
make man
```

synackd:
@cjh1 
> Hi @synackd, sorry for the delay. I don't think the use of `--uri` on the root command conflicts with using it on the service commands as well, as long as we explicitly check the value on the root command we are good.

It seems that setting `TraverseChildren` to `true` on the root command causes the flag parsing to be done top-down from the parent versus bottom-up from the child, and prevents the child `--uri` flag from completely overriding the root `--uri` flag. This means that relative paths can be passed for the service-level URI and it properly gets appended to the base cluster URI, and behavior is as predicted:

```
# Correctly tries to reach out to https://my.cluster/bss/service/status
ochami --uri 'https://my.cluster' bss status --uri '/bss'

# Correctly tries to reach out to https://different.cluster/boot/v2/service/status
ochami --uri 'https://my.cluster' bss status --uri 'https://different.cluster/boot/v2'

# Correctly tries to reach out to https://my.cluster/boot/v1/service/status
ochami --uri 'https://my.cluster' bss status

# Correctly tries to reach out to https://different.cluster/boot/v2/service/status
ochami bss status --uri 'https://different.cluster/boot/v2'

# Correctly errs with:
# failed to get base URI for BSS error="could not get bss base URI: invalid service URI for bss: bss.uri is a relative path but cluster.uri not set"
ochami -l debug bss status --uri '/boot/v2'
```

One worry I have is that, since the flags are named the same, the position that they are placed in now matters, versus before when they were named differently and their position did not matter. I could foresee this being a source of confusion when troubleshooting. What do you think?

synackd:
@davidallendj:
> > The README hasn't been updated in a while. There is #9 open to add manual page references to the README as well as build instructions. Building the manual pages uses scdoc, and, if you didn't want to use Goreleaser to build an entire release, there is a Makefile target for it:
> > ```
> > make man
> > ```
> 
> Right, I knew that, but looking at it from the perspective of someone new to the repo, it would be helpful to have it in the documentation somewhere which seems like that's being updated simultaneously in the other PR you mentioned.
> 
> Other than the stuff I mentioned documentation-wise (which seems to be handled elsewhere), I think things look good to me.

The reference to the manual pages in the README is handled in the linked PR, so that will take care of that. The help message references and manual page updates for config merging I can put in this PR.

davidallendj:
> One worry I have is that, since the flags are named the same, the position that they are placed in now matters, versus before when they were named differently and their position did not matter. I could foresee this being a source of confusion when troubleshooting. What do you think?

I think this would frustrate me especially having it as the same flag. It can also be inconvenient to have to worry about the flag position even when they're not the same, so personally I wouldn't like this at all.

cjh1:
I don't have a strong opinion, so will defer to @davidallendj.

synackd:
I think it's reasonable to keep the `--cluster-uri`/`--uri` flags separate for now. If there is reasonable demand to unify them in the future, we can address it. For now, I will address @davidallendj's documentation requests and then request a review.

cjh1:
If this is the route we are going to take we should probably use the `cluser-uri` as the key in the yaml as well, even though will look a little odd as its already under the cluster object.

synackd:
> If this is the route we are going to take we should probably use the `cluser-uri` as the key in the yaml as well, even though will look a little odd as its already under the cluster object.

Hmm, that might get a bit verbose with "cluster".

```
ochami config cluster set foobar cluster.cluster-uri https://foobar.cluster
```

cjh1:
> > If this is the route we are going to take we should probably use the `cluser-uri` as the key in the yaml as well, even though will look a little odd as its already under the cluster object.
> 
> Hmm, that might get a bit verbose with "cluster".
> 
> ```
> ochami config cluster set foobar cluster.cluster-uri https://foobar.cluster
> ```

Yep, it looks ugly, hence my thought of using uri for both. I guess its not terrible if properties in `ochami config show` don't match the cli flags, but I just thought it might confuse some ...

synackd:
Perhaps adding to the `--cluster-uri` help output that it overrides `cluster.uri` from the config file to make it clear is a good compromise?

synackd:
OK, I added:

- Help "hints" that get logged when an error occurs. (e.g: `see 'ochami config show --help' for long command help`)
- References to relevant manual pages in the long command usage output (printed with `--help`)
- More detail in the manual pages about `--cluster-uri`, `--uri`, and config key format.

synackd:
@davidallendj Could you review?

cmd/smd-component-add.go
```diff
@@ -21,31 +22,32 @@ var componentAddCmd = &cobra.Command{
...
 	Example: `  ochami smd component add x3000c1s7b56n0 56
   ochami smd component add --state Ready --enabled --role Compute --arch X86 x3000c1s7b56n0 56
   ochami smd component add -f payload.json
   ochami smd component add -f payload.yaml --payload-format yaml
   echo '<json_data>' | ochami smd component add -f -
   echo '<yaml_data>' | ochami smd component add -f - --payload-format yaml`,
-	Run: func(cmd *cobra.Command, args []string) {
+	PreRunE: func(cmd *cobra.Command, args []string) error {
 		// Check that all required args are passed
 		if len(args) == 0 && !cmd.Flag("payload").Changed {
```
davidallendj, line 36: I'm guessing you're checking the `args` here because it's a more complex check opposed to using `Args` field like the other commands?

cmd/smd-rfe-add.go
```diff
@@ -16,36 +17,38 @@ import (
 // rfeAddCmd represents the smd-rfe-add command
 var rfeAddCmd = &cobra.Command{
 	Use:   "add -f <payload_file> | (<xname> <name> <ip_addr> <mac_addr>)",
+	Args:  cobra.MaximumNArgs(4),
```
davidallendj, line 20: I'm guessing this isn't `cobra.ExactNArgs(4)` because of the optional flags?

internal/config/config.go
```diff
@@ -422,6 +571,221 @@ func DeleteConfig(path, key string) error {
...
+		return "", nil
+	}
+	switch val.(type) {
+	case map[string]interface{}, []interface{}:
+		var err error
+		var valBytes []byte
+		switch format {
+		case "yaml":
+			valBytes, err = yaml.Marshal(val)
+		case "json":
```
davidallendj, line 686: Are we supporting more than YAML for the config? I saw that some of the documentation removed references of supporting YAML, JSON, and TOML to just supporting YAML.

davidallendj: Nevermind, I misunderstood what this is doing.

davidallendj (approved):
LGTM. This is a long PR, so it's possible that I may have missed something.

internal/config/config.go
synackd: Only YAML config files are currently supported. However, different output formats when printing results are supported, which is what the linked lines are for.

synackd:
Thanks, @davidallendj.

I'm going to squash these commits into more meaningful ones for the sake of release notes and history cleanliness.

PR #13, opened 2025-03-17
add PCS transitions

cjh1:
This PR adds support for interacting with PCS transitions. The basic operations are:

```
ochami pcs transition start ...
ochami pcs transition monitor ...
ochami pcs transition show ...
ochami pcs transition list ...
ochami pcs transition abort
```

These have been tested using the [redfish emulator](https://github.com/OpenCHAMI/csm-redfish-interface-emulator), started from within [this](https://github.com/OpenCHAMI/power-control/blob/main/docker-compose.test.ct.yaml) docker compose configuration.


davidallendj:
I'm looking at this and noticed that the sub-command should be `transition` without the "s". Is there a place in the `power-control` repo with a full example for using this feature?

cjh1:
> I'm looking at this and noticed that the sub-command is `transition` without the "s". Is there a place in the `power-control` repo with a full example for using this feature?

That is a mistake in the description, although I am happy to debate if it should be `transitions`.

davidallendj:
Nah, I don't think that's a big deal.

cjh1:
Looks like the other command are using singular rather than plural ....

synackd:
I tend to go with the singular in the other commands since it tends to act somewhat as a verb. I think `transition` is fine.

pkg/client/pcs/pcs.go
```diff
@@ -83,3 +86,105 @@ func (pc *PCSClient) GetHealth() (client.HTTPEnvelope, error) {
...
+	Location     []locationEntry `json:"location"`
+}
+
+type locationEntry struct {
+	Xname     string `json:"xname"`
+}
+
+// CreateTransition is a wrapper function around OchamiClient.PostData to
+// hit the /transitions endpoint
+func (pc *PCSClient) CreateTransition(operation string, taskDeadline *int, xnames []string) (client.HTTPEnvelope, error){
```
synackd, line 102: Small nitpick: space between `(client.HTTPEnvelope, error)` and `{`

pkg/client/pcs/pcs.go
```diff
@@ -83,3 +86,105 @@ func (pc *PCSClient) GetHealth() (client.HTTPEnvelope, error) {
...
+
+	// Now create the HTTPBody
+	httpBody, err := client.BytesToHTTPBody(bytes, "json")
+	if err != nil {
+		return henv, fmt.Errorf("CreateTransition: failed to create HTTPBody: %w", err)
+	}
+
+	henv, err = pc.PostData(PCSTransitions, "", nil, httpBody)
+	if err != nil {
+		err = fmt.Errorf("CreateTransition(): error creating PCS health: %w", err)
```
synackd, line 131: Small consistency thing: `CreateTransition` -> `CreateTransition()`

pkg/client/pcs/pcs.go
```diff
@@ -83,3 +86,105 @@ func (pc *PCSClient) GetHealth() (client.HTTPEnvelope, error) {
...
+// GetTransitions is a wrapper function around OchamiClient.GetData to
+// hit the /transitions/{transitionID} endpoint
+func (pc *PCSClient) GetTransition(id string) (client.HTTPEnvelope, error) {
+	var (
+		henv              client.HTTPEnvelope
+		err               error
+		pcsTransitionsEndpoint string
+	)
+
+	pcsTransitionsEndpoint = path.Join(PCSTransitions, id)
```
synackd, line 162: I've taken to using [`url.JoinPath()`](https://pkg.go.dev/net/url#JoinPath) (see [here](https://github.com/OpenCHAMI/ochami/blob/1de633871a8354880d3a03a274e966bd0bd0dc0f/pkg/client/smd/smd.go#L710)) instead of `path.Join` since it is geared toward URI paths.

Looking around in early client code (e.g. `pkg/client/smd/smd.go`), I see that there is still string concatenation that needs to be updated with this method as well...

synackd (changes requested):
I haven't compiled and tested yet, but here is an initial code review.

cmd/lib.go
```diff
@@ -431,3 +433,23 @@ func logHelpError(cmd *cobra.Command) {
 func logHelpWarn(cmd *cobra.Command) {
 	log.Logger.Warn().Msgf("see '%s --help' for long command help", cmd.CommandPath())
 }
+
+// formatOutput formats the output as JSON or YAML
+func formatOutput(output interface{}, format string) ([]byte, error) {
+	switch strings.ToLower(format) {
+	case "json":
+		if bytes, err := json.MarshalIndent(output, "", "  "); err != nil {
```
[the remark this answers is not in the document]
cjh1: Can add in the `json-pretty` option here.

cmd/pcs-status.go
```diff
@@ -6,16 +6,20 @@ import (
...
 	"github.com/OpenCHAMI/ochami/internal/log"
 	"github.com/OpenCHAMI/ochami/pkg/client"
 	"github.com/OpenCHAMI/ochami/pkg/client/pcs"
-	"github.com/elliotchance/pie/v2"
-	"github.com/spf13/cobra"
+)
+
+const (
+	readyStatus = "ready"
+	liveStatus  = "live"
```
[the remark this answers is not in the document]
cjh1: I can add a prefix

cmd/lib.go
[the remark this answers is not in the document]
synackd: Have you taken a look at `FormatBody()`? It basically does what this function does.

cmd/pcs-transition-abort.go
```diff
@@ -0,0 +1,91 @@
...
+	Long: `Abort a PCS transition.
+
+See ochami-pcs(1) for more details.`,
+	Example: `  # Abort a transition
+  ochami pcs transition abort --id 8f252166-c53c-435e-8354-e69649537a0f`,
+	Run: func(cmd *cobra.Command, args []string) {
+		// Without a base URI, we cannot do anything
+		pcsBaseURI, err := getBaseURIPCS(cmd)
+		if err != nil {
+			log.Logger.Fatal().Err(err).Msg("failed to get base URI for PCS")
```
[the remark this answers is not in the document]
cjh1: So you want me to change this to `log.Logger.Error()` + `logHelpError(cmd)` + `os.Exit(1)` ?

cmd/pcs-transition-abort.go
```diff
@@ -0,0 +1,91 @@
...
+			log.Logger.Fatal().Msg("failed to unmarshal abort transitions response")
+		}
+
+		// Print output
+		outFmt, err := cmd.Flags().GetString("format-output")
+		if err != nil {
+			log.Logger.Fatal().Err(err).Msg("failed to get value for --format-output")
+		}
+
+		if outBytes, err := formatOutput(output, outFmt); err != nil {
```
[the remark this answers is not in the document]
cjh1: The `client.FormatBody()` version take a HTTPBody, here we need to be able handle `interface{}`, as we don't alway forward the HTTP response as the output of the command.

cmd/pcs-transition-abort.go
```diff
@@ -0,0 +1,91 @@
...
+			fmt.Println(string(outBytes))
+		}
+	},
+}
+
+func init() {
+	pcsTransitionAbortCmd.Flags().StringP("id", "i", "", "The transition ID to delete")
+	if err := pcsTransitionAbortCmd.MarkFlagRequired("id"); err != nil {
+		log.Logger.Fatal().Err(err).Msg("failed to mark id as required")
+	}
```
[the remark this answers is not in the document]
cjh1: Not really, I can do that

cmd/pcs-transition.go
```diff
@@ -0,0 +1,32 @@
...
+	Use:   "transition",
+	Args:  cobra.NoArgs,
+	Short: "Manage PCS transitions",
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) == 0 {
+			err := cmd.Usage()
+			if err != nil {
+				log.Logger.Error().Err(err).Msg("failed to print usage")
+				os.Exit(1)
+			}
```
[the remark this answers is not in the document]
cjh1: OK, will do

man/ochami-pcs.1.sc
```diff
@@ -72,6 +72,93 @@ This command accepts the following options:
...
+		Output response data in specified _format_. Supported values are:
+
+		- _json_ (default)
+		- _yaml_
+
+	*-i, --id* _id_
+		ID of the power transition to monitor.
+
+*abort*
+
```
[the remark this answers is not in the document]
cjh1: Will do

pkg/client/pcs/pcs.go
```diff
@@ -13,6 +15,7 @@ const (
 	PCSRelpathLiveness  = "/liveness"
 	PCSRelpathReadiness = "/readiness"
 	PCSRelpathHealth    = "/health"
+	PCSTransitions = "/transitions"
```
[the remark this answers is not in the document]
cjh1: will do

pkg/client/pcs/pcs.go
cjh1: `gofmt` will take care of that

pkg/client/pcs/pcs.go
cjh1: I think I got this from another service, will update it

cmd/pcs-transition-abort.go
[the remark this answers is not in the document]
synackd: All occurrences, please. It will make these easier to find when refactoring.

cmd/lib.go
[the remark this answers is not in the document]
cjh1: `FormatBody()` formats a HTTP body, this version can takes `interface{}` I can refactor things so that `FormatBody` will call my version ...

cmd/pcs-transition-monitor.go
```diff
@@ -0,0 +1,151 @@
...
+// Create and style a progress bar
+func createBar(p *mpb.Progress, name string) *mpb.Bar {
+	return p.AddBar(0, mpb.PrependDecorators(
+		decor.Name(name, decor.WC{W: 12, C: decor.DindentRight}),
+	),
+		mpb.AppendDecorators(
+			decor.Percentage(),
+		),
+	)
+}
```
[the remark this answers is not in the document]
synackd: That's fair. Not really that important to do if this is the only command using it. Just a thought for later.

cmd/pcs-transition-abort.go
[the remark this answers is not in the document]
davidallendj: I was wondering this myself when testing. Using positional arguments seem like the right way to go unless there's another way to specify the transition, which doesn't seem like it.

cmd/pcs-transition-show.go
```diff
@@ -0,0 +1,88 @@
...
+		if outBytes, err := formatOutput(output, outFmt); err != nil {
+			log.Logger.Fatal().Err(err).Msg("failed to format output")
+		} else {
+			fmt.Println(string(outBytes))
+		}
+	},
+}
+
+func init() {
+	pcsTransitionShowCmd.Flags().StringP("id", "i", "", "The transition ID to delete")
```
davidallendj, line 81: This can be changed to positional arguments similar to `abort` as mentioned before. Also, I think the description needs to be updated.

cmd/pcs-transition-monitor.go
```diff
@@ -0,0 +1,151 @@
...
+			// Sleep poll interval
+			time.Sleep(time.Duration(pollInterval) * time.Second)
+		}
+
+		p.Shutdown()
+	},
+}
+
+func init() {
+	pcsTransitionMonitorCmd.Flags().StringP("id", "i", "", "The transition ID to delete")
```
davidallendj, line 145: Required `id` can be changed to positional argument as mentioned before and description needs updating.

cmd/pcs-transition-start.go
```diff
@@ -0,0 +1,119 @@
...
+		if outBytes, err := formatOutput(output, outFmt); err != nil {
+			log.Logger.Fatal().Err(err).Msg("failed to format output")
+		} else {
+			fmt.Println(string(outBytes))
+		}
+	},
+}
+
+func init() {
+	pcsTransitionStartCmd.Flags().StringSliceP("xname", "x", []string{}, "The list of target components")
```
davidallendj, line 112: Would it make sense to use positional arguments here as well?

cmd/pcs-transition-show.go
cjh1: Done

cmd/pcs-transition-monitor.go
cjh1: Done

cmd/pcs-transition-start.go
cjh1: For the `xname` as well? We would then have two difference positional args, so the order would matter.

synackd: What if we moved the state first and had a variable number of xnames after? E.g.
```
ochami pcs transition start on xname1 xname2 ...
```
?

cjh1:
@synackd I have addressed your comments, thanks

cmd/pcs-transition-abort.go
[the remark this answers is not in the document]
cjh1: Done

cmd/pcs-transition-start.go
davidallendj: I guess it we can keep it as is. I was thinking something like:
```bash
ochami pcs transition start on x0c0s7b0n1 x0c0s7b0n0 x0c0s4b0n1
```
But I guess that wouldn't be much different than this other than have less to type:
```bash
ochami pcs transition start on --xname "x0c0s7b0n1,x0c0s7b0n0,x0c0s4b0n1"
```

davidallendj: > What if we moved the state first and had a variable number of xnames after? E.g.
> 
> ```
> ochami pcs transition start on xname1 xname2 ...
> ```
> 
> ?

@synackd Beat me to it

synackd: I think I would like to try and avoid required flags unless there is a need to specify different types of data to pass (think of `ochami bss boot params add`, where at least one of `--mac`, `--xname`, or `--nid` needs to be specified). Having the desired state as the first argument and the variable amount of required positional arguments afterward seems simpler and more intuitive.

davidallendj: The caveat I think is if there are mutually exclusives required flags like either an `id` or an `xname` being set, but that doesn't seem to be the case here.

synackd: Right. Not sure if more types were planned for this command, but that would be a case where mutually-exclusive required flags would be justifiable.

internal/utils/utils.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"gopkg.in/yaml.v3"
+	"strings"
+)
+
+// FormatOutput formats the output as JSON or YAML
+func FormatOutput(output interface{}, format string) ([]byte, error) {
```
synackd, line 11: I like the idea of separating `FormatBody` into something more generic that can be wrapped. However, I think it wold be better to put it into a more specific `format` package that is public versus a `utils` internal package that could grow indefinitely.

I went ahead and made this change in [`c772355`](https://github.com/synackd/ochami/commit/c77235545abe388531e3c7a8bff0d0e5ea11837b) on my branch [here](https://github.com/synackd/ochami/tree/refactor-transitions) for you to take a look.

synackd (changes requested):
I get the following error when trying to build the latest commit:
```
# github.com/OpenCHAMI/ochami/cmd
cmd/pcs-transition.go:6:2: "os" imported and not used
cmd/pcs-transition.go:10:2: "github.com/OpenCHAMI/ochami/internal/log" imported and not used
```

We can go ahead and get rid of those packages there since the usage printing routine was moved.

internal/utils/utils.go
davidallendj: Why make it public? Is the formatting not specific only to the CLI tool?

synackd: It's a pretty generic function with fundamental data types for inputs and outputs, plus the client packages are already public (in case one wanted to use the library functions there to communicate with OpenCHAMI services in other programs), so I figured it could be public to be used to format request data like PCS is currently doing.

cjh1: Do you want me just update the package name here?

synackd: @cjh1:
> Do you want me just update the package name here?

I already made the changes in the commit I linked, including changing the package imports and names in all of the function references.

cjh1: OK, I understand that, are you planning on raising a PR against my branch or just pushing the commits here.

cjh1: Or do you want me to cherry-pick the changes ?

synackd: Honestly, it's up to you. I was thinking you could add my remote and fast-forward merge, but cherry-picking works just as well. Then you can squash as you like into your feature commit.

cmd/pcs-transition-start.go
cjh1: > Right. Not sure if more types were planned for this command, but that would be a case where mutually-exclusive required flags would be justifiable.

I agree that using positional arguments looks clear. However, I think that plan would be to support other ways of identifying the components for the transition, for example `nid` ( we can support any primary key that smd provides ). One of the complains about the existing `cray power ` CLI is the lack of flexibility in terms of selecting the nodes to transition, so using a flag leaves open the option to implement other selection options.

davidallendj: I think if that's the intention, then leaving it as is makes sense.

davidallendj (approved):
LGTM. I tested with the changes and everything worked as expected.

cmd/pcs-transition-abort.go
```diff
@@ -0,0 +1,86 @@
...
+		if outBytes, err := format.FormatData(output, outFmt); err != nil {
+			log.Logger.Fatal().Err(err).Msg("failed to format output")
+		} else {
+			fmt.Println(string(outBytes))
+		}
+	},
+}
+
+func init() {
+	pcsTransitionAbortCmd.Flags().StringP("format-output", "F", defaultOutputFormat, "format of output printed to standard output (json,yaml)")
```
synackd, line 84: Sorry, this is probably something I should have originally done before, but could you change `(json,yaml)` to `(json,json-pretty,yaml)` in all of the `format-output` flag definitions in PCS since we added `json-pretty` as a format option?

We can change the rest in the repo after merge.

cmd/pcs-transition-start.go
```diff
@@ -0,0 +1,125 @@
...
+// pcsTransitionStartCmd represents the "pcs transition start" command
+var pcsTransitionStartCmd = &cobra.Command{
+	Use:   "start",
+	Args:  cobra.ExactArgs(1),
+	Short: "Start a PCS transition",
+	Long: `Start a PCS transition.
+
+See ochami-pcs(1) for more details.`,
+	Example: `  # Turn on a set of nodes
+  ochami pcs transition start --xname "x0c0s7b0n1,x0c0s7b0n0,x0c0s4b0n1" on`,
```
synackd, line 53: I know we and @davidallendj were talking about replacing the required flag with positional arguments. I think having the transition always as the first argument and one more more xnames to apply the transition on would be elegant.

E.g. instead of:
```
ochami pcs transition start --xname "x0c0s7b0n1,x0c0s7b0n0,x0c0s4b0n1" on
```
it would be:
```
ochami pcs transition start on x0c0s7b0n1 x0c0s7b0n0 x0c0s4b0n1
```
That way, we wouldn't need to define a flag but could leverage Cobra's argument check and parse the `args` slice.

```golang
var pcsTransitionStartCmd = &cobra.Command{
	Use:   "start <transition> <xname>...",
	Args:  cobra.MinimumNArgs(2),
	Short: "Start a PCS transition",
	Long: `Start a PCS transition.
See ochami-pcs(1) for more details.`,
	Example: `  # Turn on a set of nodes
  ochami pcs transition start on x0c0s7b0n1 x0c0s7b0n0 x0c0s4b0n1`,
```

What do you think?

man/ochami-pcs.1.sc
```diff
@@ -72,6 +73,97 @@ This command accepts the following options:
 *--vault*
 	Print out the backend vault connection status of PCS.
 
+## transitions
+
+Manages PCS transitions.
+
+Subcommands for this command are as follows:
+
+*start*  [-F _format_] [-x _xname1,xname2,..._]... _operation_
```
synackd, line 82: (Contingent on above comment)

If we change the `start` command, we'll need to change this too.

cmd/pcs-transition-start.go
synackd: Sorry, I didn't see this until just now. That makes sense @cjh1. You can ignore my comments in my most recent review. Maybe making a note for intended future support for different node IDs?

synackd: (Thank goodness for email notifications or I would have completely missed this since it was resolved.)

cmd/pcs-transition-start.go
synackd: Disregard above comment (see [comment](https://github.com/OpenCHAMI/ochami/pull/13#discussion_r2003393584)). Instead of changing the format, can we change the `Use` message to reflect the expected arguments (similar to above comments)? E.g.

```golang
var pcsTransitionStartCmd = &cobra.Command{
	Use:   "start <transition>",
	Args:  cobra.ExactArgs(1),
```

Cobra will add `[flags]` after it, so we don't need to do that ourselves.

man/ochami-pcs.1.sc
synackd: Disregard above comment (see [comment](https://github.com/OpenCHAMI/ochami/pull/13#discussion_r2003393584)).

cmd/pcs-transition-start.go
davidallendj: I agree with the change, but @cjh1 mentioned adding other flags other than just `xname` like `nid`. I don't think it makes sense to use positional arguments only to have to change it back later.

synackd: Looks like I wasn't quick enough on the draw... 😉

cjh1:
@synackd I have addressed the latest comments

synackd:
Things are looking good, but when I try to do
```
ochami pcs transition list -F yaml
```
I get a list of integers representing bytes:
```yaml
- 123
- 34
- 116
- 114
- 97
- 110
- 115
- 105
- 116
- 105
[...snip...]
- 101
- 100
- 34
- 58
- 48
- 125
- 125
- 93
- 125
```
The JSON looks fine, so it might be some conversion mishap in the formatting function(s).

cjh1:
> Things are looking good, but when I try to do
> 
> ```
> ochami pcs transition list -F yaml
> ```
> 
> I get a list of integers representing bytes:
> 
> ```yaml
> - 123
> - 34
> - 116
> - 114
> - 97
> - 110
> - 115
> - 105
> - 116
> - 105
> [...snip...]
> - 101
> - 100
> - 34
> - 58
> - 48
> - 125
> - 125
> - 93
> - 125
> ```
> 
> The JSON looks fine, so it might be some conversion mishap in the formatting function(s).

Will take a look!

synackd (approved):
Working now.

Was able to test basic functionality with the docker-compose file linked, and things worked as expected.

LGTM!

Issue #10, opened 2025-03-06
[DEV] Update `cloud-init` command to OpenCHAMI cloud-init v1 API


synackd:
# Updating Cloud-Init to v1

The OpenCHAMI [cloud-init server](https://github.com/OpenCHAMI/cloud-init) has changed considerably in v1 than the previous v0, including the API endpoints. This issue serves as the discussion for how to update the `ochami` interface to interact with this new API. The [cloud-init OpenAPI specification](https://github.com/OpenCHAMI/cloud-init/blob/main/docs/swagger.yaml) can be consulted for API documentation.

The `cloud-init` command is meant to be a "plumbing" command, along with `smd` and `bss`.

# Command Summary

- `ochami cloud-init defaults (set | get)`
- `ochami cloud-init group (add <id> | get [[--cloud-config|--meta-data] id] | set <id> | delete [id])`
- `ochami cloud-init node (get [--user-data|--vendor-data|--group <name>] [id] | set [id])`

# Command Details

### `get`

**GET** `/cloud-init/admin/cluster-defaults`

**Description:** Get cluster defaults

### `set [-f <path>] [-F <format>]`

**POST** `/cloud-init/admin/cluster-defaults`

**Description:** Set cluster defaults. Read data from stdin unless **-f** passed. JSON-formatted unless **-F** specified.

**Payload:**

```json
{
  "availability-zone": "string",
  "base-url": "http://demo.openchami.cluster:8081/cloud-init",
  "boot-subnet": "string",
  "cloud_provider": "string",
  "cluster-name": "demo",
  "nid-length": 3,
  "public-keys": [
    "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIMLtQNuzGcMDatF+YVMMkuxbX2c5v2OxWftBhEVfFb+U user1@demo-head",
    "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIB4vVRvkzmGE5PyWX2fuzJEgEfET4PRLHXCnD1uFZ8ZL user2@demo-head"
  ],
  "region": "string",
  "short-name": "nid",
  "wg-subnet": "string"
}
```

### `add [-f <path>] [-F <format>] <id>`

**POST** `/cloud-init/admin/groups`

**Description:** Add new group `<id>` with optional data. Read data from stdin unless **-f** passed. JSON-formatted unless **-F** specified.

**Payload:**

```json
{
  "description": "The compute group",
  "file": {
[...]

PR #18, opened 2025-03-24
fix: various fixes/refactoring in preparation for cloud-init changes

synackd:
These are some fixes and refactorings that were made while implementing the cloud-init changes mentioned in #10.

These include:

- Updating Makefile to properly find all source files to know when to rebuild
- Improving some of the client error messages when reading payload data
- Fixing an error when trying to marshal an HTTP body using the new `format` package
  - Includes changing `format.FormatData()` into `format.UnmarshalData()` and `format.MarshalData()`
- Changing the `--format-input` and `--format-output` flags to use an enum to better handle unknown/unsupported formats

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,12 @@ ifeq ($(SHELL),)
 $(error '$(SHELL)' command not found.)
 endif
 
+# Recursive wildcard function, obtained from https://stackoverflow.com/a/18258352
+#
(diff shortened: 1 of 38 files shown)


synackd:
I've tested all of these changes while testing cloud-init. Going to merge so we can go ahead with the cloud-init work.

PR #19, opened 2025-03-24
feat!: update cloud-init commands to match v1 API

synackd:
closes #10

Update cloud-init commands to match the cloud-init v1 API.

Since some flag changes have been added since the linked issue was created, the format of the commands in this PR is different than in the issue.

This PR also updates Go from 1.22 to 1.23.

BREAKING CHANGES: the old config/data cloud-init commands are replaced with defaults/group/node

# Command Summary

## Overview

- `ochami cloud-init defaults (set | get)`
- `ochami cloud-init group (add | delete | get (raw | config | meta-data) | render | set)`
- `ochami cloud-init node (get (group | meta-data | user-data | vendor-data))`

## Detailed

```
ochami cloud-init defaults get [OPTIONS]
ochami cloud-init defaults set [OPTIONS]
ochami cloud-init group add [OPTIONS]
ochami cloud-init group delete [OPTIONS] ([-d (<data> | @<path>)] [-f <format>]) | <group>...
ochami cloud-init group get [OPTIONS] raw [<id>...]
ochami cloud-init group get [OPTIONS] config [<id>...]
ochami cloud-init group get [OPTIONS] meta-data [id...]
ochami cloud-init group render <group> <id>
ochami cloud-init group set [OPTIONS]
ochami cloud-init node get group [OPTIONS] group <id>...
ochami cloud-init node get meta-data [OPTIONS] <id>...
ochami cloud-init node get user-data [OPTIONS] <id>...
ochami cloud-init node get vendor-data [OPTIONS] <id>...
```

# Command Details

### `get [-F <format>]`

**GET** `/cloud-init/cluster-defaults`

**Description:** Get cluster defaults. Output format in JSON unless changed with **-F**.

### `set [-f <format] [-d (<data> | @<path> | @-)]`

**POST** `/cloud-init/cluster-defaults`

**Description:** Set cluster defaults. Read data from stdin unless **-d** specifies otherwise. JSON-formatted unless **-f** specified.

**Payload:**

```json
{
  "availability-zone": "string",
  "base-url": "http://demo.openchami.cluster:8081/cloud-init",
  "boot-subnet": "string",
  "cloud_provider": "string",
  "cluster-name": "demo",
  "nid-length": 3,
  "public-keys": [
    "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIMLtQNuzGcMDatF+YVMMkuxbX2c5v2OxWftBhEVfFb+U user1@demo-head",
[...]

PR #22, opened 2025-04-08
feat: change `--force` to `--no-confirm` for clarity

synackd:
From [this comment](https://github.com/OpenCHAMI/ochami/pull/19#discussion_r2031951786), `--no-confirm` was identified as a more succinct way of telling `ochami` not to prompt the user to ask if they really want to delete something than `--force`.


davidallendj (approved):
LGTM

PR #25, opened 2025-04-09
feat!: move discover cmd to discover-static command

synackd:
This is to serve as a separate subcommand, distinguished from dynamic discovery (e.g. Magellan's subcommands) that may be added later.

Also, and importantly, the new `discover static` command uses the `[-d (<data> | @<path>)] [-f <format>]` syntax for specifying data payload, which follows the paradigm of other data modification commands. This change makes the command syntax consistent and predictable, and also allows easier specifying data from standard input and as an argument on the command line.

BREAKING CHANGE: The functionality in `discover` is now in `discover static`.


davidallendj (approved):
LGTM

PR #28, opened 2025-04-11
hotfix: add missing YAML reflect tags

synackd:
Missing YAML reflect tags caused certain fields to be blank when unmarshalling using YAML.


synackd:
Tested using `ochami discover static -f yaml -@nodes.yaml`.

PR #29, opened 2025-04-14
Update smd rfe help output and man page entry

bmcdonald3:
While working with the RedFish emulator in OpenCHAMI and trying to add some hardware with a configuration file, I noticed that there is some incorrect help information, where `-f` was used when `-d` should have been.

It was also pointed out that the man page for the `rfe` subcommand was missing, so that is added here as well.

### DIFF
--- a/man/ochami-smd.1.sc
+++ b/man/ochami-smd.1.sc
@@ -436,6 +436,141 @@ Subcommands for this command are as follows:
 		this flag can be specified multiple times or this flag can be specified
 		once and multiple xnames, separated by commas.
 
+## rfe
+
+Manage Redfish endpoints. 
+
+Subcommands for this command are as follows:
+
+*add* [--domain _domain_] [--hostname _hostname_] [--username _user_] [--password _pass_] _xname_ _name_ _ip_addr_ _mac_addr_++
+*add* [-f _format_] -d _data_++
+*add* [-f _format_] -d @_path_++
+*add* [-f _format_] -d @-++
+	Add one or more new Redfish endpoints to SMD.
+
+	In the first form of the command, an _xname_ (unique identifier), _name_
+	(human-readable name), _ip_addr_ (IP address), and _mac_addr_ (MAC address)
+	are required to identify and define the endpoint to add. Optional flags like
+	*--domain*, *--hostname*, *--username*, and *--password* can provide
+	additional details.
+
+	In the second form of the command, raw data is passed as an argument to be 
+	the payload.
+	
+	In the third form of the command, a file containing the payload data is
+	passed. This is convenient in cases of dealing with many components at once.
+
+	In the fourth form of the command, the payload data is read from standard
+	input.
+
+	This command sends a POST request to SMD's /RedfishEndpoints endpoint. An
+	access token is required.
+
+	This command accepts the following options:
+
+	*-d, --data* (_data_ | @_path_ | @-)
+		Specify raw _data_ to send, the _path_ to a file to read payload data
+		from, or to read the data from standard input (@-). The format of data
+		read in any of these forms is JSON by default unless *-f* is specified
+		to change it.
+
+	*--domain* _domain_
+		Specify the domain part of the Redfish endpoint's FQDN.
+
+	*-f, --format-input* _format_
+		Format of input payload data used by *-d*. Supported formats are:
+
+		- _json_ (default)
+		- _json-pretty_
+		- _yaml_
+
+	*--hostname* _hostname_
+		Specify the hostname part of the Redfish endpoint's FQDN.
+
+	*--password* _password_
+		Specify the password to use when interrogating the endpoint (stored in SMD).
+
+	*--username* _username_
+		Specify the username to use when interrogating the endpoint (stored in SMD).
+
+*delete* [--no-confirm] --all++
+*delete* [--no-confirm] _xname_...++
+*delete* [--no-confirm] [-f _format_] -d _data_++
+*delete* [--no-confirm] [-f _format_] -d @_path_++
+*delete* [--no-confirm] [-f _format_] -d @-++
+	Delete one or more Redfish endpoints in SMD. Unless *--no-confirm* is passed, the
+	user may be asked to confirm deletion.
+
+	In the first form of the command, all Redfish endpoints are deleted.
+
+	In the second form of the command, one or more _xname_ arguments identifying the
+	endpoint(s) to delete are specified.
+
+	In the third form of the command, a file containing the payload data is
+	passed. This is convenient in cases of dealing with many components at once.
+
+	This command sends one or more DELETE requests to SMD's /RedfishEndpoints
+	endpoint. An access token is required.
+
+	This command accepts the following options:
+
+	*-a, --all*
+		Delete *all* Redfish endpoints in SMD.
+
+	*-d, --data* (_data_ | @_path_ | @-)
+		Specify raw _data_ to send, the _path_ to a file to read payload data
+		from, or to read the data from standard input (@-). The format of data
+		read in any of these forms is JSON by default unless *-f* is specified
+		to change it.
+
+	*-f, --format-input* _format_
+		Format of input payload data used by *-d*. Supported formats are:
+
+		- _json_ (default)
+		- _json-pretty_
+		- _yaml_
+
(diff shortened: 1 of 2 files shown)


synackd:
Thank you for catching this. It also looks like the entire `rfe` command is missing from the **smd**(1) manual page. Would you be willing to add it to `man/ochami-smd.1.sc`? You can take a look at [the `component` command section](https://github.com/bmcdonald3/ochami/blob/smd-flags/man/ochami-smd.1.sc#L303-L438) as a reference for format and wording.

If not, I can take a crack at it and merge it into this branch.

bmcdonald3:
Yes, I'd be happy to, I'll put that on my list for tomorrow! Thanks.

synackd:
Thank you, that would be very helpful! That's my bad for not including it in the first place...

bmcdonald3:
OK, think this is ready for review!

man/ochami-smd.1.sc
```diff
@@ -436,6 +436,141 @@ Subcommands for this command are as follows:
...
+
+	*--password* _password_
+		Specify the password to use when interrogating the endpoint (stored in SMD).
+
+	*--username* _username_
+		Specify the username to use when interrogating the endpoint (stored in SMD).
+
+*delete* --all [--force]++
+*delete* _xname_... [--force]++
+*delete* -d (_data_ | @_path_ | @-) [-f _format_] [--force]
```
synackd, line 499: `--force` was changed to `--no-confirm` in 6932105. We should update as such.

man/ochami-smd.1.sc
```diff
@@ -436,6 +436,141 @@ Subcommands for this command are as follows:
...
+	This command accepts the following options:
+
+	*-a, --all*
+		Delete *all* Redfish endpoints in SMD.
+
+	*-d, --data* (_data_ | @_path_ | @-)
+		Specify raw _data_ payload, the _path_ to a file containing the payload
+		(prefixed with '@'), or '@-' to read the payload data from standard input.
+
+	*--force*
```
synackd, line 526: In accordance with comment above, `--force` -> `--no-confirm`, and should be moved below `-f` to be ordered alphabetically.

man/ochami-smd.1.sc
```diff
@@ -436,6 +436,141 @@ Subcommands for this command are as follows:
...
+
+*delete* --all [--force]++
+*delete* _xname_... [--force]++
+*delete* -d (_data_ | @_path_ | @-) [-f _format_] [--force]
+	Delete one or more Redfish endpoints in SMD. Unless *--force* is passed, the
+	user may be asked to confirm deletion.
+
+	In the first form of the command, all Redfish endpoints are deleted.
+
+	In the second form of the command, one or more _xname_s identifying the
```
synackd, line 505: There is a compile error on this line:

```
Error at 508:0: Expected _ before starting new paragraph (began with _ at 505:50)
```

`scdoc` won't interpret `_` surrounded by non-whitespace characters as the boundary for an underline. Thus, we'll have to find a workaround for plural, e.g by replacing `_xname_s` with `_xname_ arguments`

man/ochami-smd.1.sc
```diff
@@ -436,6 +436,141 @@ Subcommands for this command are as follows:
...
+	endpoints with complex configurations defined in the payload.
+
+	This command sends a POST request to SMD's /RedfishEndpoints endpoint. An
+	access token is required.
+
+	This command accepts the following options:
+
+	*-d, --data* (_data_ | @_path_ | @-)
+		Specify raw _data_ payload, the _path_ to a file containing the payload
+		(prefixed with '@'), or '@-' to read the payload data from standard input.
```
synackd, line 476: It's sufficient (and consistent) to just copy and paste `-d` from one of the other commands here. 😉 (And in the other occurrences.)

synackd (changes requested):
Thanks for your work in this! I just have a few requests to make things consistent with the other subcommands. (Sorry if I seem nitpicky, it's no critique of effort on your part!)

synackd:
I also noticed some small things that need updating outside of the `rfe` subcommand in this manual page I need to fix... I'll push those commits on this PR after approval.

synackd:
I see it's still a draft, but it's looking good so far. 🙂

bmcdonald3:
OK, I think it is ready to go now, I was a little less clear on the 2 that I've left unresolved, but believe I've addressed those now with the latest commit. Apologies if I'm misunderstanding there...

man/ochami-smd.1.sc
```diff
@@ -436,6 +436,146 @@ Subcommands for this command are as follows:
...
+		Specify the password to use when interrogating the endpoint (stored in SMD).
+
+	*--username* _username_
+		Specify the username to use when interrogating the endpoint (stored in SMD).
+
+*delete* [--no-confirm] --all++
+*delete* [--no-confirm] _xname_...++
+*delete* [-f _format_] [--no-confirm] -d _data_++
+*delete* [-f _format_] [--no-confirm] -d @_path_++
+*delete* [-f _format_] [--no-confirm] -d @-++
```
synackd, line 504: Sort of a nitpick, but having all of the `[--no-confirm]`s in a line would look better, I think.

man/ochami-smd.1.sc
```diff
@@ -436,6 +436,141 @@ Subcommands for this command are as follows:
...
+	are required to identify and define the endpoint to add. Optional flags like
+	*--domain*, *--hostname*, *--username*, and *--password* can provide
+	additional details.
+
+	In the second form of the command, payload data is passed using the *-d*
+	flag. The argument can be the raw _data_ itself, a _path_ to a file prefixed
+	with '@' (e.g., `@payload.json`), or '@-' to read the data from standard
+	input. The format of this data defaults to JSON but can be changed using the
+	*-f* flag. This method is convenient for adding multiple endpoints or
+	endpoints with complex configurations defined in the payload.
```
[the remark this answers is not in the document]
synackd: The goal here is to summarize each form of the command listed in the four command forms above. Since we have three different forms of using `-d`, we can omit the verbose description of general use of the `-d` flag from the second command form description (moving it to the flag description itself, which already has it) in favor of a more brief one focusing on the specific form:

> In the second form of the command, raw data is passed as an argument to be the payload.

What I meant by this comment is that this should be applied to all subcommands using `-d` as well.

man/ochami-smd.1.sc
```diff
@@ -436,6 +436,146 @@ Subcommands for this command are as follows:
...
+
+	*--no-confirm*
+		Do not ask the user to confirm deletion.
+
+	*-f, --format-input* _format_
+		Format of input payload data used by *-d*. Supported formats are:
+
+		- _json_ (default)
+		- _json-pretty_
+		- _yaml_
```
synackd, line 539: Since `--force` is now `--no-confirm`, we should put it alphabetically below `--format-input`.

man/ochami-smd.1.sc
```diff
@@ -436,6 +436,146 @@ Subcommands for this command are as follows:
...
+	In the first form of the command, all Redfish endpoints are deleted.
+
+	In the second form of the command, one or more _xname_ arguments identifying the
+	endpoint(s) to delete are specified.
+
+	In the third form of the command, payload data identifying the endpoints to
+	delete is passed using the *-d* flag. The argument can be the raw _data_
+	itself, a _path_ to a file prefixed with '@' (e.g., `@payload.json`), or
+	'@-' to read the data from standard input. The format of this data defaults
+	to JSON but can be changed using the *-f* flag.
```
synackd, line 517: Similar to my comment above, this block should describe specifically what the third form of the command does and omit the description of the different argument types to `-d`. That sort of detailed description should be put into the flag description below (which it already is).

man/ochami-smd.1.sc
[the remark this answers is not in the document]
synackd: Looks good!

synackd:
Looks good! Is it okay if I squash these commits and reword the commit message to conform to [conventional commits](https://www.conventionalcommits.org)?

bmcdonald3:
Great, thank you for the feedback and review! Yes, no problem on the squashing and rewording. I will read up on conventional commits... Thanks again!

synackd (approved):

PR #31, opened 2025-06-02
chore: update cloud-init in go.mod to v1.2.3

synackd:
To incorporate YAML marshalling fixes from:

https://github.com/OpenCHAMI/cloud-init/pull/79


synackd:
I've tested this and confirmed it is working as expected.

PR #34, opened 2025-06-25
Add flag to make second request to SMD optional to prevent 409 conflicts creating `EthernetInterface`s

davidallendj:
This PR removes the second request to SMD to create the `EthernetInterface`s, which is not necessary since SMD now creates the interfaces with the discovery output included in the `RedfishEndpoint` data.

One thing to note here is that the `SchemaVersion` has no impact on whether the `EthernetInterface`s are added to SMD. Instead, the `SchemaVersion` determines whether to use the old versus current output from Magellan (see [note](https://github.com/OpenCHAMI/magellan?tab=readme-ov-file#openchami-magellan) in README).

### DIFF
--- a/cmd/discover-static.go
+++ b/cmd/discover-static.go
@@ -92,7 +92,7 @@ See ochami-discover(1) for more details.`,
 
 		// Put together payload for different endpoints
 		log.Logger.Debug().Msg("generating redfish structures to send to SMD")
-		comps, rfes, _, err := discover.DiscoveryInfoV2(smdBaseURI, nodes)
+		comps, rfes, ifaces, err := discover.DiscoveryInfoV2(smdBaseURI, nodes)
 		if err != nil {
 			log.Logger.Error().Err(err).Msg("failed to construct structures to send to SMD")
 			logHelpError(cmd)
@@ -239,87 +239,93 @@ See ochami-discover(1) for more details.`,
 		}
 
 		// Send EthernetInterface requests
-		// var (
-		// 	ifaceErrorsOccurred bool = false
-		// 	ifaceHenvs          []client.HTTPEnvelope
-		// 	ifaceErrs           []error
-		// 	ifaceErr            error
-		// )
-		// if cmd.Flag("overwrite").Changed {
-		// 	// SMD's EthernetInterface API does not allow the PUT
-		// 	// method. Instead, we loop over each ethernet interface
-		// 	// to add and attempt a POST. Iff a 409 is returned for
-		// 	// that interface, a PATCH is attempted. Otherwise, an
-		// 	// error has occurred.
-		// 	for _, iface := range ifaces {
-		// 		// Attempt to POST the ethernet interface
-		// 		ifaceListWrapper := []smd.EthernetInterface{iface}
-		// 		ifaceHenvs, ifaceErrs, ifaceErr = smdClient.PostEthernetInterfaces(ifaceListWrapper, token)
+		var (
+			ifaceErrorsOccurred bool = false // could also just be...ifaceErr != nil || len(ifaceErrs) > 0
+			ifaceHenvs          []client.HTTPEnvelope
+			ifaceErrs           []error
+			ifaceErr            error
+		)
+		discoveryVersion, err := cmd.Flags().GetInt("discovery-version")
+		if err != nil {
+			log.Logger.Warn().Err(err).Msgf("could not get discovery version...using default value of %d", discoveryVersion)
+		}
+		if discoveryVersion == 1 {
+			if cmd.Flag("overwrite").Changed {
+				// SMD's EthernetInterface API does not allow the PUT
+				// method. Instead, we loop over each ethernet interface
+				// to add and attempt a POST. If a 409 is returned for
+				// that interface, a PATCH is attempted. Otherwise, an
+				// error has occurred.
+				for _, iface := range ifaces {
+					// Attempt to POST the ethernet interface
+					ifaceListWrapper := []smd.EthernetInterface{iface}
+					ifaceHenvs, ifaceErrs, ifaceErr = smdClient.PostEthernetInterfaces(ifaceListWrapper, token)
 
-		// 		if ifaceErr != nil {
-		// 			// An error in the function occurred, err for
-		// 			// this interface and move on.
-		// 			log.Logger.Error().Err(ifaceErr).Msg("failed to add ethernet interface to SMD")
-		// 			ifaceErrorsOccurred = true
-		// 			continue
-		// 		}
+					if ifaceErr != nil {
+						// An error in the function occurred, err for
+						// this interface and move on.
+						log.Logger.Error().Err(ifaceErr).Msg("failed to add ethernet interface to SMD")
+						ifaceErrorsOccurred = true
+						continue
+					}
 
-		// 		if ifaceErrs[0] != nil {
-		// 			// An HTTP error occurred
-		// 			if errors.Is(ifaceErrs[0], client.UnsuccessfulHTTPError) {
-		// 				if ifaceHenvs[0].StatusCode == 409 {
-		// 					// Ethernet interface exists, patch it
-		// 					log.Logger.Info().Msgf("ethernet interface with MAC address %s exists, attempting to update it", iface.MACAddress)
-		// 					_, patchErrs, patchErr := smdClient.PatchEthernetInterfaces(ifaceListWrapper, token)
-		// 					if patchErr != nil {
-		// 						log.Logger.Error().Err(patchErr).Msg("failed to update existing ethernet interface in SMD")
-		// 						ifaceErrorsOccurred = true
-		// 						continue
-		// 					}
-		// 					if patchErrs[0] != nil {
-		// 						var errMsg string
-		// 						if errors.Is(patchErrs[0], client.UnsuccessfulHTTPError) {
-		// 							errMsg = "SMD ethernet interface PATCH yielded unsuccessful HTTP response"
-		// 						} else {
-		// 							errMsg = "failed to update existing ethernet interface in SMD"
-		// 						}
-		// 						log.Logger.Error().Err(patchErrs[0]).Msg(errMsg)
-		// 						ifaceErrorsOccurred = true
-		// 						continue
-		// 					}
-		// 				} else {
-		// 					// Some other HTTP error occurred, err
-		// 					log.Logger.Error().Err(ifaceErrs[0]).Msg("SMD ethernet interface POST yield non-409 (duplicate) failure")
-		// 					ifaceErrorsOccurred = true
-		// 					continue
-		// 				}
-		// 			} else {
-		// 				log.Logger.Error().Err(ifaceErrs[0]).Msg("failed to add ethernet interface to SMD")
-		// 				ifaceErrorsOccurred = true
-		// 				continue
-		// 			}
-		// 		}
-		// 	}
-		// } else {
-		// 	// --overwrite was not passed, perform regular POST.
-		// 	_, ifaceErrs, ifaceErr = smdClient.PostEthernetInterfaces(ifaces, token)
-		// 	if ifaceErr != nil {
-		// 		log.Logger.Error().Err(ifaceErr).Msg("failed to add ethernet interfaces to SMD")
-		// 		ifaceErrorsOccurred = true
-		// 	}
-		// 	for _, err := range ifaceErrs {
-		// 		if err != nil {
-		// 			var errMsg string
-		// 			if errors.Is(err, client.UnsuccessfulHTTPError) {
-		// 				errMsg = "SMD ethernet interface request yielded unsuccessful HTTP response"
-		// 			} else {
-		// 				errMsg = "failed to add ethernet interface to SMD"
-		// 			}
-		// 			log.Logger.Error().Err(err).Msg(errMsg)
-		// 			ifaceErrorsOccurred = true
-		// 		}
-		// 	}
-		// }
+					if ifaceErrs[0] != nil {
+						// An HTTP error occurred
+						if errors.Is(ifaceErrs[0], client.UnsuccessfulHTTPError) {
+							if ifaceHenvs[0].StatusCode == 409 {
+								// Ethernet interface exists, patch it
+								log.Logger.Info().Msgf("ethernet interface with MAC address %s exists, attempting to update it", iface.MACAddress)
+								_, patchErrs, patchErr := smdClient.PatchEthernetInterfaces(ifaceListWrapper, token)
+								if patchErr != nil {
+									log.Logger.Error().Err(patchErr).Msg("failed to update existing ethernet interface in SMD")
+									ifaceErrorsOccurred = true
+									continue
+								}
+								if patchErrs[0] != nil {
+									var errMsg string
+									if errors.Is(patchErrs[0], client.UnsuccessfulHTTPError) {
+										errMsg = "SMD ethernet interface PATCH yielded unsuccessful HTTP response"
+									} else {
+										errMsg = "failed to update existing ethernet interface in SMD"
+									}
+									log.Logger.Error().Err(patchErrs[0]).Msg(errMsg)
+									ifaceErrorsOccurred = true
+									continue
+								}
+							} else {
+								// Some other HTTP error occurred, err
+								log.Logger.Error().Err(ifaceErrs[0]).Msg("SMD ethernet interface POST yield non-409 (duplicate) failure")
+								ifaceErrorsOccurred = true
+								continue
+							}
+						} else {
+							log.Logger.Error().Err(ifaceErrs[0]).Msg("failed to add ethernet interface to SMD")
+							ifaceErrorsOccurred = true
+							continue
+						}
+					}
+				}
+			} else {
+				// --overwrite was not passed, perform regular POST.
+				_, ifaceErrs, ifaceErr = smdClient.PostEthernetInterfaces(ifaces, token)
+				if ifaceErr != nil {
+					log.Logger.Error().Err(ifaceErr).Msg("failed to add ethernet interfaces to SMD")
+					ifaceErrorsOccurred = true
+				}
+				for _, err := range ifaceErrs {
+					if err != nil {
+						var errMsg string
+						if errors.Is(err, client.UnsuccessfulHTTPError) {
+							errMsg = "SMD ethernet interface request yielded unsuccessful HTTP response"
+						} else {
+							errMsg = "failed to add ethernet interface to SMD"
+						}
+						log.Logger.Error().Err(err).Msg(errMsg)
+						ifaceErrorsOccurred = true
+					}
+				}
+			}
+		}
 
 		// Put together list of groups to add and which components to add to those groups
 		groupsToAdd := make(map[string]smd.Group)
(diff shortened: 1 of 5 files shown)


synackd:
In that case, I'm wondering if it is worth adding a compatibility flag in order to use `ochami` with older SMD versions. Any feedback on this @alexlovelltroy ?

synackd:
Even if the old way is deprecated in a later release, I think it might be a good idea to keep a way to do the old way around for a bit with sufficient warning that it will be deprecated later.

man/ochami-discover.1.sc
```diff
@@ -91,6 +91,11 @@ RedfishEndpoints, EthernetInterfaces, Components, and groups data in SMD
 corresponding to each node. It also creates Components corresponding to each
 node's BMC which corresponds to each RedfishEndpoint created.
 
+The *--discovery-version* sets which discovery method to use when running the
+*static* subcommand. If the version is set the 1, an additional request is made
+to create the EthernetInterfaces separately in SMD. Otherwise, the EthernetInterfaces
+are created with the first discovery request.
+
```
synackd, line 98: > Otherwise, the EthernetInterfaces are created with the first discovery request.

I think we'll want to be explicit about the versioning, in the chance that it increases again. I think we should also be explicit about the new version being 2 (and it being the default).

```suggestion
The *--discovery-version* sets which discovery method to use when running the
*static* subcommand. If the version is set to 1, an additional request is made
to create the EthernetInterfaces separately in SMD. If set to 2 (the default),
the EthernetInterfaces are created with the first discovery request.

This flag is only for backward compatibility with earlier versions of SMD and
may be deprecated in a later version of ochami.

```

man/ochami-discover.1.sc
```diff
@@ -110,6 +115,9 @@ This command accepts the following options:
 	Instead of failing if data already exists, overwrite it with new data
 	contained in the payload.
 
+*--discovery-version*
+  Set the version of the discovery method to use for static discovery.
+
```
synackd, line 120: We should include the possible values here.

```suggestion
*--discovery-version*
        Set the version of the discovery method to use for static discovery.
   
        Possible values are:

        - _1_
        - _2_ (default)
```

man/ochami-discover.1.sc
```diff
@@ -110,6 +115,9 @@ This command accepts the following options:
 	Instead of failing if data already exists, overwrite it with new data
 	contained in the payload.
 
+*--discovery-version*
+  Set the version of the discovery method to use for static discovery.
```
synackd, line 119: `scdoc` requires tabs for indentation.

cmd/discover-static.go
```diff
@@ -239,84 +239,90 @@ See ochami-discover(1) for more details.`,
...
+							errMsg = "SMD ethernet interface request yielded unsuccessful HTTP response"
+						} else {
+							errMsg = "failed to add ethernet interface to SMD"
+						}
+						log.Logger.Error().Err(err).Msg(errMsg)
+						ifaceErrorsOccurred = true
+					}
+				}
 			}
 		}
```
synackd, line 327: I think we'll want to check if the discovery version is not `1` or `2` here.

synackd (changes requested):
Finished with my initial review. I think it might make sense to define constants for discovery version values so we aren't writing integers in multiple places. E.g:

```go
const (
        DISCOVER_V1 int = iota + 1      // 1
        DISCOVER_V2                     // 2
)
```

It builds successfully, but I still need to test.

cmd/discover-static.go
```diff
@@ -239,84 +239,90 @@ See ochami-discover(1) for more details.`,
 
 		// Send EthernetInterface requests
 		var (
-			ifaceErrorsOccurred bool = false
+			ifaceErrorsOccurred bool = false // could also just be...ifaceErr != nil || len(ifaceErrs) > 0
```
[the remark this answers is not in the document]
davidallendj: Could you not just check for `len(ifaceErrs) < 0` then as opposed to storing with another variable entirely? I added the `ifaceErr != nil` part because `ifaceErrorsOccurred=true` when `ifaceErr` is non-nil here:
https://github.com/OpenCHAMI/ochami/blob/allend/fix-conflict-error/cmd/discover-static.go#L267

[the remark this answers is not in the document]
synackd: No, because `nil` values get added to the slice if an error did not occur for that request.

cmd/discover-static.go
```diff
@@ -451,6 +457,7 @@ See ochami-discover(1) for more details.`,
 }
 
 func init() {
+	discoverStaticCmd.Flags().Int("discovery-version", 1, "set version for discovery method to use")
```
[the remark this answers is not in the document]
davidallendj: I thought you mentioned `discovery-version` being 1 was the preferred in your comment above?

cmd/discover-static.go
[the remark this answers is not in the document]
synackd: The reason for this implementation is so that the indexes of the error slice line up with the indexes of the HTTP response slice so that errors can be matched to which request/response emitted them.

cmd/discover-static.go
davidallendj: Actually, I think where it's at now is fine but I have to change to `discoveryVersion == 2` instead. I think I was zero-indexing the version number to match how it's done in SMD.

synackd: I do think we should check for a version that's not `1` or `2` and err if not.

man/ochami-discover.1.sc
synackd: Whoops, that's my bad on the indentation here. It should be:

```scdoc
*--discovery-version*
        Set the version of the discovery method to use for static discovery.

        Possible values are:
        - _1_
        - _2_ (default)
```

man/ochami-discover.1.sc
```diff
@@ -22,24 +22,24 @@ format is as follows:
...
+		- name: internal
+			ip_addr: 172.16.0.1
+	- mac_addr: de:ad:be:ee:ee:f2
+		ip_addrs:
+		- name: external
+			ip_addr: 10.15.3.100
+	- mac_addr: 02:00:00:91:31:b3
+		ip_addrs:
+		- name: HSN
+			ip_addr: 192.168.0.1
```
synackd, line 42: We'll want to keep these as spaces to have nicely-formatted YAML.

man/ochami-discover.1.sc
```diff
@@ -91,6 +91,12 @@ RedfishEndpoints, EthernetInterfaces, Components, and groups data in SMD
 corresponding to each node. It also creates Components corresponding to each
 node's BMC which corresponds to each RedfishEndpoint created.
 
+The *--discovery-version* sets which discovery method to use when running the
+*static* subcommand. If the version is set to 1, an additional request is made
+to create the EthernetInterfaces separately in SMD. If set to 2 (the default),
+the EthernetInterfaces are created with the first discovery request.
+This flag is only for backward compatibility with earlier versions of SMD and
+may be deprecated in a later version of ochami.
 This command accepts the following options:
```
synackd, line 100: Can we add a new line between this line and the previous line? E.g:

```scdoc
The *--discovery-version* sets which discovery method to use when running the
*static* subcommand. If the version is set to 1, an additional request is made
to create the EthernetInterfaces separately in SMD. If set to 2 (the default),
the EthernetInterfaces are created with the first discovery request.
This flag is only for backward compatibility with earlier versions of SMD and
may be deprecated in a later version of ochami.

This command accepts the following options:
```

All of the other times "This command accepts the following options:" appears, it's on its own line.

man/ochami-discover.1.sc
```diff
@@ -110,6 +116,14 @@ This command accepts the following options:
...
 	contained in the payload.
 
+*--discovery-version*
+	Set the version of the discovery method to use for static discovery.
+	Set the version of the discovery method to use for static discovery.
+	 
+				Possible values are:
+				- _1_
+				- _2_ (default)
+
```
synackd, line 126: Whoops, that's my bad on the indentation here. It should be:

```scdoc
*--discovery-version*
        Set the version of the discovery method to use for static discovery.

        Possible values are:
        - _1_
        - _2_ (default)
```

man/ochami-discover.1.sc
synackd: Gonna resolve this and address new changes in a different review.

synackd (changes requested):
Just a couple of other minor changes. I'll test this tomorrow.

cmd/discover-static.go
davidallendj: Sure, but maybe it would be better to check the value of the argument early in the process with `cobra.Args` and have it failed there instead of here, before anything is done. What do you think?

man/ochami-discover.1.sc
davidallendj: I think this may have been change automatically by VSCode when saving. Is that 4 spaces or tab spaces?

man/ochami-discover.1.sc
davidallendj: I think this one if fixed with https://github.com/OpenCHAMI/ochami/pull/34/commits/0d0606694d11fa1670ad199680e9b48713ef3cfd.

cmd/discover-static.go
davidallendj: I added a check with https://github.com/OpenCHAMI/ochami/pull/34/commits/d0b260f0269cf64b0d03adf1ab168a58946e47d8 in `cmd.Args` that checks if the value is set to either `1` or `2`. Otherwise, it returns an error stating that the value must be one of the valid values.

man/ochami-discover.1.sc
davidallendj: Is it just the YAML section or the entire document for spacing? I'm assuming it's just the YAML.

man/ochami-discover.1.sc
synackd: Also can we move this block to be between `-d` and `-f` so the flags are in alphabetical order?

davidallendj:
Can you just send me a patch to apply formatting?

synackd:
> Can you just send me a patch to apply formatting?

```diff
diff --git a/man/ochami-discover.1.sc b/man/ochami-discover.1.sc
index ee62226..b3037c6 100644
--- a/man/ochami-discover.1.sc
+++ b/man/ochami-discover.1.sc
@@ -118,12 +118,11 @@ This command accepts the following options:
 	contained in the payload.
 
 *--discovery-version*
-			Set the version of the discovery method to use for static discovery.
-			Set the version of the discovery method to use for static discovery.
-	 
-				Possible values are:
-				- _1_
-				- _2_ (default)
+	Set the version of the discovery method to use for static discovery.
+
+	Possible values are:
+	- _1_
+	- _2_ (default)
 
 # XNAMES
 
```

man/ochami-discover.1.sc
synackd: Looks good now

synackd:
Things seem in order, I think the only thing left is [this comment](https://github.com/OpenCHAMI/ochami/pull/34#discussion_r2180903423).

pkg/discover/version.go
```diff
@@ -0,0 +1,45 @@
...
+
+type DiscoveryVersion int
+
+const (
+	DiscoveryMethodV1 DiscoveryVersion = iota + 1 // 1
+	DiscoveryMethodV2                             // 2
+)
+
+var (
+	DisocveryVersionHelp = map[int]string{
```
synackd, line 16: `DisocveryVersionHelp` -> `DiscoveryVersionHelp`

cmd/discover-static.go
```diff
@@ -16,8 +16,21 @@ import (
...
+			return err
+		}
+		// check that value of discovery-version is valid or else...
+		switch discoveryVersion {
+		case int(discover.DiscoveryMethodV1), int(discover.DiscoveryMethodV2):
+			return cobra.NoArgs(cmd, args)
+		default:
+			return fmt.Errorf("--discovery-version value must be 1 or 2")
+		}
+	},
```
synackd, line 33: The `--discovery-version` enum should handle the checks so we shouldn't need this anymore.

cmd/discover-static.go
```diff
@@ -244,79 +257,83 @@ See ochami-discover(1) for more details.`,
...
-			// to add and attempt a POST. Iff a 409 is returned for
-			// that interface, a PATCH is attempted. Otherwise, an
-			// error has occurred.
-			for _, iface := range ifaces {
-				// Attempt to POST the ethernet interface
-				ifaceListWrapper := []smd.EthernetInterface{iface}
-				ifaceHenvs, ifaceErrs, ifaceErr = smdClient.PostEthernetInterfaces(ifaceListWrapper, token)
+		// get discovery version value (err handled in cmd.Args)
+		discoveryVersion, _ := cmd.Flags().GetInt("discovery-version")
+		if discoveryVersion == 2 {
```
synackd, line 262: Instead of `2`, we should use `discover.DiscoveryMethodV2` here.

cmd/discover-static.go
```diff
@@ -451,6 +468,7 @@ See ochami-discover(1) for more details.`,
 }
 
 func init() {
+	discoverStaticCmd.Flags().Var(&discoveryVersion, "discovery-version", "set version for discovery method to use")
```
synackd, line 471: A few lines below this is (`cloudInitGroupAddCmd` needs to be changed to `discoverStaticCmd`, I can do that in a subsequent PR):

```go
cloudInitGroupAddCmd.RegisterFlagCompletionFunc("format-input", completionFormatData)
```

which sets the shell completion function for the `--format-input` enum flag. We should add a line below this one to set the function for the discovery version:

```go
discoverStaticCmd.RegisterFlagCompletionFunc("discovery-version", completionDiscoveryVersion)
```

Then, we need to implement the `completionDiscoveryVersion` function similar to this one:

https://github.com/OpenCHAMI/ochami/blob/78a2b046518839bbd8283804905e1648dd739927/cmd/lib.go#L445-L451

Go ahead and put it below it. Here's the implementation (make sure the `github.com/OpenCHAMI/ochami/pkg/discover` package is imported in `cmd/lib.go`):

```go
// completionDiscoveryVersion is the cobra completion function for the
// --discovery-version flag.
func completionDiscoveryVersion(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
	var helpSlice []string
	for k, v := range discover.DiscoveryVersionHelp {
		helpSlice = append(helpSlice, fmt.Sprintf("%s\t%s", k, v))
	}
	return helpSlice, cobra.ShellCompDirectiveDefault
}
```

cmd/discover-static.go
```diff
@@ -244,79 +257,83 @@ See ochami-discover(1) for more details.`,
...
-			// method. Instead, we loop over each ethernet interface
-			// to add and attempt a POST. Iff a 409 is returned for
-			// that interface, a PATCH is attempted. Otherwise, an
-			// error has occurred.
-			for _, iface := range ifaces {
-				// Attempt to POST the ethernet interface
-				ifaceListWrapper := []smd.EthernetInterface{iface}
-				ifaceHenvs, ifaceErrs, ifaceErr = smdClient.PostEthernetInterfaces(ifaceListWrapper, token)
+		// get discovery version value (err handled in cmd.Args)
+		discoveryVersion, _ := cmd.Flags().GetInt("discovery-version")
```
synackd, line 261: We can delete this line. The `discoveryVersion` variable is set with the flag definition below (line 471) with the default being set in `cmd/root.go` on line 27.

pkg/discover/version.go
```diff
@@ -0,0 +1,45 @@
...
+		int(DiscoveryMethodV1): "One-line JSON format",
+		int(DiscoveryMethodV2): "Unindented JSON format",
```
synackd, line 18: Copy-paste error for the descriptions? These should describe the discovery versions instead of the JSON formats.

pkg/discover/version.go
```diff
@@ -0,0 +1,45 @@
...
+	default:
+		return fmt.Errorf("must be one of %v", []DiscoveryVersion{
+			DiscoveryMethodV1,
+			DiscoveryMethodV2,
+		})
+	}
+}
+
+func (dv DiscoveryVersion) Type() string {
+	return "DiscoveryVersion "
```
synackd, line 44: Extra trailing space after `DiscoveryVersion`.

```suggestion
	return "DiscoveryVersion"
```

synackd (changes requested):

cmd/discover-static.go
synackd: Implemented in 9e5fd0b

cmd/discover-static.go
davidallendj: Removed in https://github.com/OpenCHAMI/ochami/pull/34/commits/79cce53fb055718fb0ff04c9992ef53f735deabc.

cmd/discover-static.go
davidallendj: Fixed in https://github.com/OpenCHAMI/ochami/pull/34/commits/79cce53fb055718fb0ff04c9992ef53f735deabc.

cmd/discover-static.go
davidallendj: Added in https://github.com/OpenCHAMI/ochami/pull/34/commits/79cce53fb055718fb0ff04c9992ef53f735deabc.

pkg/discover/version.go
davidallendj: Fixed in https://github.com/OpenCHAMI/ochami/pull/34/commits/79cce53fb055718fb0ff04c9992ef53f735deabc.

synackd:
Resolved all conversations. @davidallendj could you rebase this branch onto `main` so that the automated checks will run? We'll also need to incorporate the new `--discovery-version` flag into the tests.

synackd:
Cool, looks like all of the checks are passing now. Do you have a method you've been testing this with?

pkg/discover/version.go
```diff
@@ -0,0 +1,45 @@
...
+
+var (
+	DiscoveryVersionHelp = map[int]string{
+		int(DiscoveryMethodV1): "Discovery with additional request to add EthernetInterfaces",
+		int(DiscoveryMethodV2): "Discovery without request to add EthernetInterfaces",
+	}
+)
+
+func (dv DiscoveryVersion) String() string {
+	return fmt.Sprintf("v%d", (dv))
```
synackd, line 23: We should get rid of the `v` prefix here. It causes the help output for the flag to be:
```
--discovery-version DiscoveryVersion   set version for discovery method to use (default v2)
```
Notice the `v2`. As a user, I'm tempted to use that as the argument to the flag via `--discovery-version v2`, which yields:

```
Error: invalid argument "v2" for "--discovery-version" flag: strconv.Atoi: parsing "v2": invalid syntax
```

synackd (changes requested):
I tested this with SMD v2.18.0 with both discovery version 1 and 2 which yielded the same results (success, no errors). What is the expected result here? I presume version 1 should fail (return 409) with this version of SMD while version 2 should succeed.

davidallendj:
> I tested this with SMD v2.18.0 with both discovery version 1 and 2 which yielded the same results (success, no errors). What is the expected result here? I presume version 1 should fail (return 409) with this version of SMD while version 2 should succeed.

No, both should succeed. SMD v2.18.0 should be able to accept both versions. The only difference is that the EthernetInterfaces are created upon discovery vs being created in a second request. Both versions should be valid with that version. Using an earlier version should also succeed, but shouldn't create the EthernetInterfaces with the discovery output.

synackd:
Testing with v2.18.0 with both versions, I can see the EthernetInterfaces for the Node type get created with v2, and not v1.

v1:
```
$ ./ochami discover static -f yaml -d @/etc/openchami/data/nodes.yaml --log-level debug --discovery-version 1
$ ochami smd iface get | jq '.[] | select(.Type=="NodeBMC")' | wc -l
78
$ ochami smd iface get | jq '.[] | select(.Type=="Node")' | wc -l
0
```
v2:
```
$ ./ochami discover static -f yaml -d @/etc/openchami/data/nodes.yaml --log-level debug --discovery-version 2
$ ochami smd iface get | jq '.[] | select(.Type=="NodeBMC")' | wc -l
78
$ ochami smd iface get | jq '.[] | select(.Type=="Node")' | wc -l
78
```

Looks like it's working properly. Once the requested change gets made, I think this is ready for approval.

synackd (approved):
LGTM

PR #38, opened 2025-07-14
test: add initial unit tests

synackd:
Add unit tests for:
- `pkg/`
  - `client/` (`http.go` and `client.go`)
  - `discover/`
  - `format/`
  - `xname/`
- `internal/`
  - `config/`
  - `io/`
  - `log/`
  - `version/`

Also, fix some issues caught by the unit tests.

Finally, add `.github/workflows/unit_tests.yml` to run unit tests on PRs and pushes to `main`.


synackd:
Even though all the checks are passing, we should probably do some manual testing just as a sanity check. The config/log code was heavily refactored so we should make sure that this didn't break expected behavior (and that the unit tests enforce this expected behavior).

cmd/lib.go
```diff
@@ -26,20 +29,80 @@ var (
...
-	fmt.Fprintf(os.Stderr, fstr+"\n", arg...)
+// loopYesNo takes prompt p and appends " [yN]: " to it and prompts the user for
+// input. As long as the user's input is not "y" or "n" (case insensitive), the
+// function redisplays the prompt. If the user's response is "y", true is
+// returned. If the user's response is "n", false is returned.
+func (i ioStream) loopYesNo(p string) (bool, error) {
+	s := bufio.NewScanner(i.stdin)
+
+	for {
+		fmt.Fprint(i.stderr, fmt.Sprintf("%s [yN]:", p))
```
davidallendj, line 91: Shouldn't "N" indicate that no is the default response? For example, if I press Enter with no input, then the response should be the same as inputting "n".

synackd: Ah, gotta adjust the tests for this change. Will do that shortly.

synackd:
Tests passing with committed fixes. Going to squash, make sure tests pass, then merge.

PR #42, opened 2025-08-01
feat(config): add --no-confirm flag to automatically create config files

adrianreber:
Add --no-confirm persistent flag to config command that automatically creates config files and parent directories without prompting for confirmation.

Implementation details:
- Add --no-confirm flag to config command as persistent flag
- Update askToCreate function to accept noConfirm parameter
- When noConfirm is true, askToCreate returns true without prompting
- Simplify handleFileCreation to pass flag directly to askToCreate
- Centralize all noConfirm logic within askToCreate function
- Maintain backward compatibility with existing askToCreate behavior
- Add test case for noConfirm functionality

🤖 Generated with [Claude Code](https://claude.ai/code)

This PR was required by lines from the OpenHPC OpenCHAMI recipe:
```shell
$ yes | ochami config cluster set --system --default ${compute_prefix} cluster.uri https://${dns_name}:8443
```
being able to run
```shell
$ ochami config --no-confirm cluster set --system --default ${compute_prefix} cluster.uri https://${dns_name}:8443
```

looks nicer.

### DIFF
--- a/cmd/config-cluster-set.go
+++ b/cmd/config-cluster-set.go
@@ -61,23 +61,8 @@ See ochami-config(5) for details on the configuration options.`,
 			fileToModify = config.UserConfigFile
 		}
 
-		// Ask to create file if it doesn't exist
-		if create, err := ios.askToCreate(fileToModify); err != nil {
-			if err != FileExistsError {
-				log.Logger.Error().Err(err).Msg("error asking to create file")
-				logHelpError(cmd)
-				os.Exit(1)
-			}
-		} else if create {
-			if err := createIfNotExists(fileToModify); err != nil {
-				log.Logger.Error().Err(err).Msg("error creating file")
-				logHelpError(cmd)
-				os.Exit(1)
-			}
-		} else {
-			log.Logger.Error().Msg("user declined to create file, not modifying")
-			os.Exit(0)
-		}
+		// Handle file creation based on --no-confirm flag
+		handleFileCreation(cmd, fileToModify)
 
 		// Perform modification
 		dflt, err := cmd.Flags().GetBool("default")
--- a/cmd/config-set.go
+++ b/cmd/config-set.go
@@ -58,23 +58,8 @@ See ochami-config(5) for details on the configuration options.`,
 			os.Exit(1)
 		}
 
-		// Ask to create file if it doesn't exist.
-		if create, err := ios.askToCreate(fileToModify); err != nil {
-			if err != FileExistsError {
-				log.Logger.Error().Err(err).Msg("error asking to create file")
-				logHelpError(cmd)
-				os.Exit(1)
-			}
-		} else if create {
-			if err := createIfNotExists(fileToModify); err != nil {
-				log.Logger.Error().Err(err).Msg("error creating file")
-				logHelpError(cmd)
-				os.Exit(1)
-			}
-		} else {
-			log.Logger.Error().Msg("user declined to create file, not modifying")
-			os.Exit(0)
-		}
+		// Handle file creation based on --no-confirm flag
+		handleFileCreation(cmd, fileToModify)
 
 		// Perform modification
 		if err := config.ModifyConfig(fileToModify, args[0], args[1]); err != nil {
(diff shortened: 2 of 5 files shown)


davidallendj:
Seems like a nice little feature. What happens if you run the same command with a config that already exists?

synackd:
@adrianreber Thank you very much for your contribution! I agree that having a way to non-interactively create the config file(s) will be useful for scripting environments as you've stated. However, we should think carefully about how we do this.

Currently, there is already [a `--no-confirm` flag](https://github.com/OpenCHAMI/ochami/blob/da5eac1d8a4ecea8df54d2b48bcee27356486951/cmd/bss-boot-params-delete.go#L168) for commands that delete things which deletes without prompting. IMO, we could re-use this flag for `config` commands. The only question would be, as stated already, should this always overwrite existing config file(s) or only if one doesn't exist? We could potentially solve this with a corresponding `--confirm` flag. That way, `--no-confirm` would automatically not create if existing (and throw an error if trying to write to a file that doesn't exist) while `--confirm` would.

On this, my thoughts are these:

- When executing commands that create/modify (`config set` ,  `config cluster set`), the user likely wants the operation to be idempotent, creating the file if non-existent or modifying any existing file.
  - Both `--confirm` and `--no-confirm` might make sense here. If the file doesn't exist with `--no-confirm`, an error would be produced.
- When executing commands that show (`config [cluster] show`), the user likely wants to see the merged config (default) unless `--config`, `--system`, or `--user` are passed. It doesn't really make sense to create a config file if any of those flags are passed.
  - The only exception I could foresee is if the user wanted to bootstrap a config file at any of those paths using the default config, but I think it would be better to have a separate command (e.g. `bootstrap`) to do this or just have the user do `ochami config show | tee <file>`.
  - Both `--confirm` and `--no-confirm` probably do not make sense here.
- When executing commands that delete (`config [cluster] unset`), the user likely expects the file to already exist and so it wouldn't make sense to create a file in this case.
  - Unless the user wants to create the file only to unset something, I do not think `--confirm` and `--no-confirm` make sense here.

Long story short, it probably only makes sense to have confirm flags on `config [cluster] set` commands.

synackd:
Apologies, the following is sort of me thinking in writing, but hopefully it helps direct the discussion. 🙂 

You bring up a good point and as I reconsider how to handle this, I realize that the current decision path for the prompting of the config file creation actually differs from that of the deletion commands.

When `--no-confirm` was written for the deletion commands, it was intended for the purpose you describe: "don't prompt, simply delete" (so conversely, a `--confirm` prompt, if implemented, would _force_ the prompt). So the decision paths for deletion are currently:

- Prompt user for deletion (no flag or hypothetical `--confirm` flag)
- Do not prompt and delete (`--no-confirm`)

However, for config file creation, I think that the expected decision paths should be:

- Do not prompt and create if non-existent 
- Do not prompt and err if non-existent (is this realistic?)
- Prompt if non-existent

Ideally, I would like to re-use flags as much as possible to avoid having different flags that do identical/similar things. That said, I think we can rework both the config flag prompting mechanism and deletion prompting mechanism via the following generalized paths:

- Prompt (default behavior; no flag passed or `-p|--prompt`)
- Non-interactively perform "yes" option (`-Y|--yes`)
- Non-interactively perform "no" option (`-N|--no` [`-n` is taken for `--nid` (should we change?), so using upper case here])

Applying this to the deletion example:

- Prompt for deletion (default behavior or `--prompt` passed)
- Non-interactively ("force") delete (`--yes` passed)
- Non-interactively do not delete (`--no` passed, though this is probably useless so probably doesn't need to be implemented)

Applying this to the config creation example:

- Prompt to create if non-existent (default behavior or `--prompt` passed)
- Non-interactively create if non-existent (`--yes` passed)
- Non-interactively do not create and err if non-existent (`--no` passed)

This allows us to reuse the same flags for commands that require yes/no prompts, subtracting as necessary, and (hopefully) reduces confusion.

What are folks' thoughts on this? If this is feasible, we can change the config options in this PR and open a subsequent one for the deletion commands.

alexlovelltroy:
Thinking about existing tools I've used.

`--force` generally means "do it instead of asking"
`--non-interactive` generally means "suppress all prompts.  No one will see them"

How can we make the flag and the behavior least surprising?

PR #44, opened 2025-08-14
feat: add `--no-token` flag and `cluster.enable-auth` config option

### LINKED ISSUE #41: [FEATURE] add a flag to skip tokens
**Is your feature request related to a problem? Please describe.**
In development environments, I often won't bother with setting up authentication, and let my services happily accept unauthenticated requests. `ochami` doesn't think this should be possible, and demands a valid, unexpired JWT. I then fiddle with some online JWT generators to produce an acceptable garbage token.

**Describe the solution you'd like**
Add a `--no-token` or `--disable-token` (do we have an informal convention on what word disable flags use? probably not, IDK) boolean flag. When enabled, `ochami` does not check for tokens and does not include them in requests.

**Describe alternatives you've considered**
Continue performing the generator dance. That suffices and is a one-off task during a dev session. This is very much a nice to have, not required feature.

**Additional context**

I'd looked at trying to do this quickly and found that it was non-trivial because it's enforced with per-command implementations (e.g. https://github.com/OpenCHAMI/ochami/blob/v0.3.5/cmd/bss.go#L27-L31 and https://github.com/OpenCHAMI/ochami/blob/v0.3.5/cmd/pcs.go#L27-L31) and toggled per subcommand (e.g. the `true` in https://github.com/OpenCHAMI/ochami/blob/v0.3.5/cmd/pcs-transition-show.go#L32).

We'll have quite a lot of commands that require a token, but also many that won't, and we should see if there's some DRYable way of writing.

synackd:
Resolves #41

Apologies for the wait on this. I had to fight the internal config-reading mechanism since adding a config option whose default value is "true" turned out to be non-trivial. Along the way, I refactored and even fixed a few things. See the extended description in some of the individual commit messages below for more details.

## Features

- Add `--no-token` flag to disable token reading/checking
- Add per-cluster `cluster.enable-auth` key (defaults to true) to persistently toggle on/off token reading/checking for individual clusters

## Changes

- Refactored token checking to happen in separate `handleToken()` function so it's not coupled to `*GetClient()` functions.
  - It's a bit more repetitive, but allows more flexibility for when tokens are checked (or if they are)
- Reworked the config method to use the YAML library's unmarshal methods so that the `enable-auth` key would be handled correctly via the custom `UnmarshalYAML`
  - Koanf's unmarshal functions use mapstructure under the hood, which won't call the `UnmarshalYAML` function of the struct its unmarshaling into
- Updated manual pages with new flag and config option
- Added `GetCluster()` function to config library to easily retrieve clusters from config by name
- Fixed type conversion of string config options to relevant types by adding `StringToType()` function

### DIFF
--- a/cmd/bss-boot-image-set.go
+++ b/cmd/bss-boot-image-set.go
@@ -33,7 +33,10 @@ See ochami-bss(1) for more details.`,
   ochami bss boot image set --mac 00:de:ad:be:ef:00,de:ca:fc:0f:fe:ee live:https://172.16.0.254/image.squashfs`,
 	Run: func(cmd *cobra.Command, args []string) {
 		// Create client to use for requests
-		bssClient := bssGetClient(cmd, true)
+		bssClient := bssGetClient(cmd)
+
+		// Handle token for this command
+		handleToken(cmd)
 
 		// Get current kernel command line args
 		values := url.Values{}
--- a/cmd/lib.go
+++ b/cmd/lib.go
@@ -369,15 +369,47 @@ func getBaseURI(cmd *cobra.Command, serviceName config.ServiceName) (string, err
 // performs any other setup tasks for tokens. It is called by all commands that
 // require a token.
 func handleToken(cmd *cobra.Command) {
-	if noToken, err := cmd.Flags().GetBool("no-token"); err != nil {
-		log.Logger.Error().Err(err).Msg("unable to set --no-token")
-		logHelpError(cmd)
-		os.Exit(1)
-	} else if !noToken {
-		setToken(cmd)
-		checkToken(cmd)
+	if cmd.Flag("no-token").Changed {
+		// --no-token overrides any cluster settings
+		log.Logger.Debug().Msg("--no-token passed, not reading or checking for token")
 	} else {
-		log.Logger.Debug().Msg("skipping token setup since --no-token passed")
+		// Check if enable-auth is set for cluster and only read/check
+		// token if true
+		var clusterName string
+		if cmd.Flag("cluster").Changed {
+			// Use cluster passed via --cluster
+			clusterName = cmd.Flag("cluster").Value.String()
+		} else if config.GlobalConfig.DefaultCluster != "" {
+			// Use default cluster
+			clusterName = config.GlobalConfig.DefaultCluster
+		}
+
+		if clusterName != "" {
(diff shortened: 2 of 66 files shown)


synackd:
Looks like you have the extra `redondo` cluster in your merged config. That config likely lives in `/etc/ochami/config.yaml`. If you add the `--user` flag to your `config show` command, you should get the expected result.

cmd/bss-boot-image-set.go
```diff
@@ -33,7 +33,10 @@ See ochami-bss(1) for more details.`,
   ochami bss boot image set --mac 00:de:ad:be:ef:00,de:ca:fc:0f:fe:ee live:https://172.16.0.254/image.squashfs`,
 	Run: func(cmd *cobra.Command, args []string) {
 		// Create client to use for requests
-		bssClient := bssGetClient(cmd, true)
+		bssClient := bssGetClient(cmd)
+
+		// Handle token for this command
+		handleToken(cmd)
```
davidallendj, line 39: Could you change the name of `handleToken` to be a bit more descriptive? Judging by how it's used, it seems like we're including/excluding the tokens in the headers and bypassing the check so maybe something like `checkToken`, `verifyToken`, or something like that? It's not immediately obvious what's happening with the current name.

cmd/lib.go
```diff
@@ -365,7 +365,55 @@ func getBaseURI(cmd *cobra.Command, serviceName config.ServiceName) (string, err
...
+		}
+
+		if clusterName != "" {
+			if cl, err := config.GlobalConfig.GetCluster(clusterName); err != nil {
+				if errors.Is(err, config.ErrUnknownCluster{}) {
+					// Cluster was not found (this error
+					// should be caught before this function, but
+					// this check is here just in case),
+					// skip token check
+					log.Logger.Warn().Msgf("cluster %q not found, not checking token", clusterName)
```
davidallendj, line 394: Should this be `%q` or `%s` here?

cmd/bss-boot-image-set.go
synackd: `handleToken` is a wrapper function defined in `lib.go` that conditionally calls `setToken()` and `checkToken()`. Since it's a wrapper that determines whether to set/check the token, the "handleX" seems appropriate since it follows the convention of the existing `handlePayload()` function.

davidallendj (approved):
Looks good

PR #45, opened 2025-08-15
hotfix: spacing issue in ochami-discover(1)

synackd:
Using spaces instead of tabs for indentation in the **ochami-discover**(1) manual page was causing `scdoc` to fail. Somehow, this was not caught by Goreleaser during the release process. Thus, an extra PR check has been added that builds the manual pages with `make man` to act as a fallback check.


alexlovelltroy (approved):
LGTM

synackd:
One sec, I need to fix the runner for these. Thanks for the review!

PR #47, opened 2025-08-20
feat: add fake power actions

rainest:
Add the System.Actions field and populate it with all possible actions during fake discovery. This allows PCS to attempt those actions.

Fake discovery does not directly expose the System struct, so there wasn't an existing place to slot this into config. We'll probably need to revisit that eventually: there are many other fields in SMD/schemas types that aren't yet in the fake discovery type, and I expect there are more unexpected behaviors when they have zero values.

Providing a permissive default on the SMD side is probably good enough for fake discovery, since you can just ignore (actually) unsupported actions by not requesting PCS perform them.

Downstream, this looks like:

1. PCS's SMD client [pulls allowable actions](https://github.com/OpenCHAMI/power-control/blob/53e048ba0dd603071a17a223eec7b706dac607dd/internal/hsm/state_manager.go#L444-L453) from the [associated SMD Redfish info type](https://github.com/OpenCHAMI/smd/blob/f07680ffb6c41e75945bc32bc7ba948a56afe2e5/pkg/sm/endpoints.go#L125-L129).
1. The power status component tracker [reworks this into an intermediate type](https://github.com/OpenCHAMI/power-control/blob/main/internal/domain/power-status.go#L328-L373), transforming Redfish power actions into PCS transition types.
1. The transition runner [rejects transitions if they're not in the supported list](https://github.com/OpenCHAMI/power-control/blob/53e048ba0dd603071a17a223eec7b706dac607dd/internal/domain/transitions.go#L1349-L1364).


pkg/discover/discover.go
```diff
@@ -153,7 +153,6 @@ func DiscoveryInfoV2(baseURI string, nl NodeList) (smd.ComponentSlice, smd.Redfi
 		systemMap := make(map[string]string)
 		managerMap := make(map[string]string)
 
-		// Create fake BMC "System" for node if it doesn't already exist
```
synackd, line 156: Any reason we are removing this comment? I don't see any change that renders it invalid.

pkg/client/smd/smd.go
```diff
@@ -90,13 +90,21 @@ type RedfishEndpointV2 struct {
...
 
+// This struct is a partially duplicated copy of the schemas InventoryDetail type, originally from SMD:
+// https://github.com/OpenCHAMI/schemas/blob/9aad17a286c405e1d75463298ec8db553cc4ca12/schemas/inventory.go#L61-L88
+// smd.parseRedfishEndpointDataV2 unmarshals the Systems field of a RedfishEndpoint request into one:
+// https://github.com/OpenCHAMI/smd/blob/f07680ffb6c41e75945bc32bc7ba948a56afe2e5/cmd/smd/smd-api.go#L2826-L2842
+// SMD will have zero values for any fields this currently lacks, and will have unexpected behavior when it tries to
+// use them later.
+
 // System represents data that would be retrieved from BMC System data, except
 // reduced to a minimum needed for discovery.
```
synackd, line 101: The merge of https://github.com/OpenCHAMI/smd/pull/74 and https://github.com/OpenCHAMI/ochami/pull/48 will require the first part of this comment to be revised. The definition will be [here](https://github.com/OpenCHAMI/smd/blob/444da15c1e0e5a296cd6282e09a5ab9d9123291f/pkg/schemas/inventory.go#L61) (this is a branch so the URI will need to be changed when merged).

Also, this comment block should be connected to the struct comment below it to indicate that the comment describes the struct. While we're at it, could we limit the line width to 80 characters to match the other comment lines? E.g.

```go
// System represents data that would be retrieved from BMC System data, except
// reduced to a minimum needed for discovery. It is a partially duplicated copy
// of ...
```

synackd (changes requested):
I think this looks good, just need to update some comments. I also still need to test this.

pkg/discover/discover.go
rainest: I'd originally added some notes to the comment, and should have reverted it rather than deleting the line, but I suppose I may as well ask about the reason for the comment. We currently have:

```
systemMap := make(map[string]string)
managerMap := make(map[string]string)

if _, ok := systemMap[node.Xname]; !ok { ... }
if _, ok := managerMap[bmcXname]; !ok { ... }
```

so those will _never_ exist in the map already, correct? That hints at maybe refactoring moving stuff around, or some intended deduplication (for nodes that share a BMC xname) that should declare the maps elsewhere.

Not related to this PR though, and if it is causing unintended behavior I didn't notice it, so I just restored the comment.

pkg/client/smd/smd.go
rainest: On a second read I think you maybe just meant the package path.

To clarify, will we still have a separate `ochami/pkg/client/smd.System` shadow type, or will ochami CLI code use `smd/v2/pkg/schemas.InventoryDetail` directly? If it's the latter we don't need this comment at all.

If it's the former we should still have it, and ideally use a tag instead of a commit. Do you know how soon we plan to have a release with https://github.com/OpenCHAMI/smd/pull/74 ?

I'd intentionally kept it separate to not include it in the docstring, since it is, for lack of better description, messy--it's halfway to being a TODO to stop using the shadow type if we keep having to duplicate fields into it.

rainest:
> Do you have instructions I can test this with? Are you using a PCS quadlet or similar deployment?

The simple option is we simply ignore the E2E testing here, it's just aligning `discover static` ComponentEndpoints with ones created by Magellan!

IDK how we want to handle that more broadly for this repo--I wouldn't want it to be responsible for E2E tests anyway, but it could maybe do single-service tests for "the actual service API accepts this request". We'd need to build out more SMD mocks for things that aren't SMD though.

The E2E answer is sorta yes, in that I have something that works, but that it's still in not release-ready form, and this is one of the chicken-egg problem pieces to getting it there. I can invite you to my sandbox repo. The PCS image currently used by that has hacks to deal with https://github.com/OpenCHAMI/ochami/pull/51 and to avoid using Vault.

pkg/client/smd/smd.go
synackd: > On a second read I think you maybe just meant the package path.

Yep.

> To clarify, will we still have a separate `ochami/pkg/client/smd.System` shadow type, or will ochami CLI code use `smd/v2/pkg/schemas.InventoryDetail` directly? If it's the latter we don't need this comment at all.
> 
> If it's the former we should still have it, and ideally use a tag instead of a commit.

The goal is to move entirely to a package within SMD and move away from the schemas repo completely. That way, SMD owns its own schema and no shadow types are needed. The shadow types were created because a while back it looked like we might transition everything to a schemas repo and the current schemas repo lacked a lot of things the ochami CLI needed, but the paradigm has since shifted.

> Do you know how soon we plan to have a release with [OpenCHAMI/smd#74](https://github.com/OpenCHAMI/smd/pull/74) ?

I don't really have an accurate gauge. I think we should merge this soon, so I'm inclined to agree that we can keep a TODO comment around for switching over to the SMD package once it gets merged, at which point in time we can get rid of the SMD shadow types.

> I'd intentionally kept it separate to not include it in the docstring, since it is, for lack of better description, messy--it's halfway to being a TODO to stop using the shadow type if we keep having to duplicate fields into it.

That's fair. Since a transition to an SMD client package is underway, I don't think this should be a blocker. If you want to add the TODO and then comment here (so I'm notified) and resolve this, we can move forward with this PR.

rainest:
Force-pushed and squashed the review fixes to get past a phantom conflict (I do love when git says there's a conflict for lines that no longer exist in the changeset and forces you to fix it twice before squashing makes it irrelevant :neutral_face: )

synackd:
In my environment, I haven't been able to test performing transitions in PCS after the nodes are discovered, but I can verify that they are being sent:
```
2025-09-18T15:51:12-06:00 INF http.go:124 > Response status: HTTP/2.0 201 Created
2025-09-18T15:51:12-06:00 DBG client.go:245 > POST: https://demo.openchami.cluster:8443/hsm/v2/Inventory/RedfishEndpoints
2025-09-18T15:51:12-06:00 DBG client.go:266 > Request headers:
2025-09-18T15:51:12-06:00 DBG client.go:268 >   User-Agent: [ochami/v0.5.3]
2025-09-18T15:51:12-06:00 DBG client.go:268 >   Authorization: [Bearer eyJhbGciOiJSUzI1NiIsImtpZCI6IjdiMTlhY2E4LWNjMTAtNDA1ZC1hYzkyLTdkYTMwYTllY2U4ZSIsInR5cCI6IkpXVCJ9.eyJhdWQiOltdLCJjbGllbnRfaWQiOiJjY2YxZGYxYS01YWU0LTQ3NGQtYjlmYS05NzdhMzA0ZTRlNzYiLCJleHAiOjE3NTgyMzU4NjYsImV4dCI6e30sImlhdCI6MTc1ODIzMjI2NiwiaXNzIjoiaHR0cHM6Ly9yZWRvbmRvLnVzcmMiLCJqdGkiOiJlNGEzYjY4Yi1iODY0LTQ2YmEtOWFmNi1jN2NhYmQwZTQyZDEiLCJuYmYiOjE3NTgyMzIyNjYsInNjcCI6WyJvcGVuaWQiLCJzbWQucmVhZCJdLCJzdWIiOiJjY2YxZGYxYS01YWU0LTQ3NGQtYjlmYS05NzdhMzA0ZTRlNzYifQ.nOeTQYY42tr_QJlGjXZuUgF5Cc1vD5Pd835PPnnasv1TcVEIwRf3R9iMo22tQAJOdsU0DO1YYooxwBLGk53doiVvJKuxkhAkdo2OD4bHW09L4QjHuzddKJO320D_v_6cYiQSXa3cCpaztF9RVinvXrVKgc5W-IZHSdemTIo2GOsnSlFroQHFwphidTrxYEglwwmrj1ct-heCz8p5e-3H-sGd0GArRpbsVw1Xz2ewPvaozJ25qLAOKM8I2QEBysemrSM3udf8nu_u8unvXxYaR3EGMskmyKM_-eHvHR3aWcsj6Lojji1Ojv1X-sItDaz5lJcNn0SNt2aG6QU7KXonZT6GBrcnFdxvhfNNqBInV3ELZ-SG3sCKxJarRFQzCsY2OSlKRiRtB8mSZcMNNWhBws0VS0tY8V7GU4qjSwcL8o-nPg70HnQy8_TEHpB1DHsb4HErf2cOTb0OXr5l5YHwWkXsm0DELLBOOTJaDdvLUwWJuCYgDrD2uvUAEpY-vbsR5UMrquw3YV46K7on-EhalxL25TsMsQUXYtUQbzaQqTAkElJkSfSndoWUxSqwlQtSV0tDjCRGByXVFJp784kM4rwnj3BZQssiztGB8Ftmw_ZOtzhPvWL_z3ZNqr9dEe_-4xsOdl5mGROWPs172L6pkSkabVjka0WgrgQtW9eXeEo]
2025-09-18T15:51:12-06:00 DBG client.go:274 > Request body:
2025-09-18T15:51:12-06:00 DBG client.go:275 > {"ID":"x3001c1s7b70","Type":"NodeBMC","Name":"re99","UUID":"4f5d0f05-a42c-4f29-a8ec-2dd48e3d4db3","MACAddr":"d0:94:66:6b:48:22","IPAddress":"172.16.0.199","DiscoveryInfo":{"LastAttempt":"0001-01-01T00:00:00Z"},"SchemaVersion":1,"Systems":[{"uri":"https://demo.openchami.cluster:8443/redfish/v1/Systems/x3001c1s7b70n0","uuid":"dfbf86d2-4e22-4277-b22f-2443b12452d4","name":"re99","ethernet_interfaces":[{"mac":"d0:94:66:6b:48:28","ip":"172.16.0.99","name":"x3001c1s7b70n0","description":"Interface 0 for re99"}],"actions":["On","ForceOff","GracefulShutdown","GracefulRestart","ForceRestart","Nmi","ForceOn","PushPowerButton","PowerCycle","Suspend","Pause","Resume"]}],"Managers":[{"uri":"https://demo.openchami.cluster:8443/redfish/v1/Managers/x3001c1s7b70","uuid":"4f5d0f05-a42c-4f29-a8ec-2dd48e3d4db3","name":"x3001c1s7b70","ethernet_interfaces":[{"mac":"d0:94:66:6b:48:22","ip":"172.16.0.199","name":"x3001c1s7b70","description":"Interface for BMC x3001c1s7b70"}],"actions":null,"description":"","type":"NodeBMC"}]}
2025-09-18T15:51:12-06:00 DBG client.go:288 > Response status: 201 Created
2025-09-18T15:51:12-06:00 DBG client.go:290 > Response headers:
2025-09-18T15:51:12-06:00 DBG client.go:292 >   Content-Type: [application/json]
2025-09-18T15:51:12-06:00 DBG client.go:292 >   Location: [/hsm/v2/Inventory/RedfishEndpoints/x3001c1s7b70]
2025-09-18T15:51:12-06:00 DBG client.go:292 >   Date: [Thu, 18 Sep 2025 21:51:12 GMT]
2025-09-18T15:51:12-06:00 DBG client.go:292 >   Content-Length: [60]
2025-09-18T15:51:12-06:00 DBG client.go:305 > Response body:
2025-09-18T15:51:12-06:00 DBG client.go:306 > [{"URI":"/hsm/v2/Inventory/RedfishEndpoints/x3001c1s7b70"}]

2025-09-18T15:51:12-06:00 INF http.go:124 > Response status: HTTP/2.0 201 Created
```

@rainest I don't want complications in my environment to be a blocker. Have you tested this to work on your end?

rainest:
It did indeed actually reboot the machine after in mine.

In the strictest sense the supported list check happens only at transition creation time, and actually sending them happens after. That's kinda pedantic, but you should be able to check that PCS doesn't accept any transition creates without this, and that it does accept them after (even if the BMC's not reachable).

synackd (approved):
I see that the transitions get added to SMD:
```
hmsds=> SELECT * FROM comp_endpoints where id='x3000c0s0b0n0';
-[ RECORD 1 ]---+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
id              | x3000c0s0b0n0
type            | Node
domain          | 
redfish_type    | ComputerSystem
redfish_subtype | 
rf_endpoint_id  | x3000c0s0b0
mac             | 
uuid            | e3da88d9-993b-4e51-bd97-c278e9f33ef6
odata_id        | /redfish/v1/Systems/x3000c0s0b0n0
component_info  | {"Name":"re01","Actions":{"#ComputerSystem.Reset":{"ResetType@Redfish.AllowableValues":["On","ForceOff","GracefulShutdown","GracefulRestart","ForceRestart","Nmi","ForceOn","PushPowerButton","PowerCycle","Suspend","Pause","Resume"],"@Redfish.ActionInfo":"/redfish/v1/Systems/x3000c0s0b0n0/ResetActionInfo","target":"/redfish/v1/Systems/x3000c0s0b0n0/Actions/ComputerSystem.Reset"}},"EthernetNICInfo":[{"RedfishId":"","@odata.id":"","Description":"Interface 0 for re01","InterfaceEnabled":true,"MACAddress":"5c:ed:8c:21:ac:08"}]}
```
and @rainest has confirmed this has worked in his setup. This LGTM and rather than delay this any further waiting for my own PCS setup to be complete, I stamp this with my approval.

Issue #55, opened 2025-10-07
[BUG] Adding a BMC that already exists causes an error


alexlovelltroy:
`ochami discover static` apparently generates the same set of post operations for every listed node.  In cases where there is one BMC with more than one node attached, the format of `nodes.yaml` requires us to include identical BMC information twice, as subkeys of two nodes.  When the code processes the BMC for a second time, SMD rejects the POST with a HTTP header to indicate a resource conflict.  Updating this behavior to suppress the failure and proceed when trying to add the BMC a second time (possibly with a log message) would make it easier to handle clusters in which multiple nodes share a single BMC.

shunr-hpe:
I'm not opposed to changing the structure of nodes.yaml.

If nodes.yaml changes, ochami will still need to post a redfish endpoint with multiple systems. So my changes shouldn't be to far out of the way of how it will be once nodes.yaml is redesigned.

synackd:
> That yaml format looks reasonable to me.
> 
> We are interested in having a fix sooner than later. I created this [PR ](https://github.com/OpenCHAMI/ochami/pull/62) to just fix the bug part.

Thanks for the feedback. Having a deduplication map seems like a reasonable fix to ship immediately while the format update is worked on.

> With regard to the new yaml format does ochami need to be backward compatible in this case? i.e. does it need to support both the old and the new yaml?

That's a good question. Should we deprecate the old format and provide a way to convert, removing the old format code in a later version?

PR #62, opened 2025-11-04
Added support for multiple nodes per BMC

shunr-hpe:
There is hardware that has multiple nodes per BMC. The code did not support this. It would create two RedfishEndpoints, each with one node. Instead it needed to create one RedfishEndpoint with two nodes.

## Pull Request Template

### Checklist

- [x] **REUSE Compliance**:  
  - [ ] Each new/modified source file has SPDX copyright and license headers  
  - [ ] Any non-commentable files include a `<filename>.license` sidecar  
  - [ ] All referenced licenses are present in the `LICENSES/` directory  

### Description

Please include a summary of the change and which issue is fixed.  
Also include relevant motivation and context.

Fixes #([55](https://github.com/OpenCHAMI/ochami/issues/55)) - This change fixes the bug part of issue 55. There is on going discussion of changing the yaml schema for static discovery.

### Type of Change

- [x] Bug fix  
- [ ] New feature  
- [ ] Breaking change  
- [ ] Documentation update  

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,6 +68,13 @@ all: binaries
 .PHONY: binaries
 binaries: $(NAME)
 
+.PHONY: unittest
+unittest:
+ifeq ($(GO),)
+	$(error go command not found.)
+endif
+	$(GO) test -cover -v ./...
+
 .PHONY: clean
 clean:
 ifeq ($(GO),)
--- a/pkg/discover/discover.go
+++ b/pkg/discover/discover.go
@@ -115,9 +115,11 @@ func DiscoveryInfoV2(baseURI string, nl NodeList) (smd.ComponentSlice, smd.Redfi
 	}
 
 	var (
-		compMap    = make(map[string]string) // Deduplication map for SMD Components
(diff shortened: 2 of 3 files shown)


synackd:
I don't see any testing procedure in the comments or PR description, but I presume this was tested before merging?

rainest:
I'm usually okay with code tests passing--ultimately, if it's new, and it didn't break existing tests and passes new ones, onus is on the author if what they merged doesn't actually handle the new functionality they need.

Alternately, new features that don't work perfectly as intended are okay (especially since there's not yet a tagged release with it yet)--you can always commit new fixes over top.

This didn't look like it was likely to break existing stuff, and I expect that if @shunr-hpe realizes this doesn't do what it should when working with pre-release builds, they'll have incentive to fix it because they're the one that need the feature.

erl-hpe (commented):
This looks right to me. We have tested that this resolves the problems that vTDS was having with registering more than a single node with the same BMC and it works for that, which unblocks vTDS moving forward with single-blade systems.

Issue #35, opened 2025-07-07
[FEATURE] Add ability to query group memberships of a node


synackd:
**Is your feature request related to a problem? Please describe.**
Not a problem, but a lack of functionality.

**Describe the solution you'd like**
It's currently possible to view the list of nodes for a group, but it would be useful to go the reverse and view the groups that a node is part of.

The SMD API already supports this, so it should be simple to implement:

```
$ curl -s -H "Authorization: Bearer $DEMO_ACCESS_TOKEN" https://demo.openchami.cluster:8443/hsm/v2/memberships/x1000c0s0b0n0 | jq
{
  "id": "x1000c0s0b0n0",
  "groupLabels": [
    "all",
    "compute",
    "slurm"
  ],
  "partitionName": ""
}
```

**Describe alternatives you've considered**
Not having this, but having it has been suggested multiple times.

**Additional context**
N/A

synackd:
Completed with the merge of #92.

PR #92, opened 2026-05-27
feature(smd): add membership subcommand

Wafffle77:
## Pull Request Template

### Checklist

- [ ] I have added tests that prove my fix is effective or my feature works  
- [X] **REUSE Compliance**:  

### Description

Adds the `ochami smd membership <node>` subcommand, which can be used to query for group memberships of a specific node.

Implements #35

### Type of Change

- [ ] Bug fix  
- [X] New feature  
- [ ] Breaking change  
- [ ] Documentation update  

---


Wafffle77:
I'm not 100% sure if my updates to the man page are style-conformant, so I'd appreciate a proofread.

cmd/smd/group/membership.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/OpenCHAMI/ochami/internal/cli"
+	"github.com/OpenCHAMI/ochami/internal/log"
+	"github.com/OpenCHAMI/ochami/pkg/client"
+
+	smd_lib "github.com/OpenCHAMI/ochami/internal/cli/smd"
+)
+
+func newCmdGroupMembership() *cobra.Command {
+	// groupGetCmd represents the "smd group get" command
```
synackd: Assumedly:
```go
// groupMembershipCmd represents the "smd group membership" command
```

cmd/smd/group/membership.go
```diff
@@ -0,0 +1,73 @@
...
+	var groupMembershipCmd = &cobra.Command{
+		Use:   "membership <node>",
+		Args:  cobra.MaximumNArgs(1),
```
synackd: Could we allow this command to get the membership of multiple nodes?

cmd/smd/group/membership.go
```diff
@@ -0,0 +1,73 @@
...
+		Short: "Get all group memberships of a node",
+		Long: `Get all group memberships of a node.
+
+See ochami-smd(1) for more details.`,
+		Example: `  ochami smd group get x1000c0s0b0n0`,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 1 {
+				return fmt.Errorf("expected 1 argument (node name), got %d", len(args))
```
synackd: This would technically be the node xname instead of the node name.

man/ochami-smd.1.sc
```diff
@@ -812,6 +812,25 @@ Subcommands for this command are as follows:
 
 This command is DEPRECATED. Use *service status* instead.
 
+## membership
+
+Get the groups a node is a member of.
+
+*membership* <node>
```
synackd:
```
*membership* _xname_
```

or even:

```
*membership* _node_xname_
```

The convention for placeholder args in these manual pages has been to underline them (surround with underscores).

synackd (changes requested):
This works nicely! Just a few small comments below.

Also, I could see it being useful to get the memberships of multiple nodes, which is probably most efficiently done by using query parameters, e.g:

GET `/hsm/v2/memberships?id=x3000c0s0b0n0&id=x3001c1s7b70n0`
to get memberships for x3000c0s0b0n0 and x3001c1s7b70n0

Which could be done via:

```
ochami smd group membership x3000c0s0b0n0 x3001c1s7b70n0
```

I could also see the use in getting all memberships if no arguments are specified.

The SMD `/memberships` API has a lot more query parameters that could be used as well. [[1]](https://github.com/OpenCHAMI/smd/blob/a4b6fa8c94678c47ffbdb3226b55ccd13e3d6735/api/swagger_v2.yaml#L5585-L5601) [[2]](https://github.com/OpenCHAMI/smd/blob/a4b6fa8c94678c47ffbdb3226b55ccd13e3d6735/api/swagger_v2.yaml#L12118) (You can copy the contents of [swagger_v2.yaml](https://github.com/OpenCHAMI/smd/blob/main/api/swagger_v2.yaml) into an OpenAPI editor such as [editor.swagger.io](https://editor.swagger.io/) to interactively look through the API spec.)

As a reference/hint for query parameters, take a look at [the `smd rfe get` command](https://github.com/OpenCHAMI/ochami/blob/main/cmd/smd/rfe/get.go) which calls [`GetRedfishEndpoints()`](https://github.com/OpenCHAMI/ochami/blob/bb8dc458d8521c0927a2dd6942d2a2f217e43d45/pkg/client/smd/smd.go#L236) which in turn calls [`GetData()`](https://github.com/OpenCHAMI/ochami/blob/bb8dc458d8521c0927a2dd6942d2a2f217e43d45/pkg/client/client.go#L112). You essentially craft the query parameters using a [`url.Values`](https://pkg.go.dev/net/url#Values) and then pass it as a string to your get function.

synackd:
Also, there were two new required CI checks that were added to satisfy #91, which has been merged. If you rebase onto the current `main`, these checks will be satisfied.

man/ochami-smd.1.sc
Wafffle77: Should be resolved in 77b650e71e4e203419a1740d7f3ab9d273cc94fc

cmd/smd/group/membership.go
Wafffle77: Should be resolved in 3df9405f4d98ea02104af3f54dd1c82c3e8eb566 (This argument got factored out)

cmd/smd/group/membership.go
Wafffle77: Fixed in 3df9405f4d98ea02104af3f54dd1c82c3e8eb566. All options are passed via flags that can be specified multiple times.

cmd/smd/group/membership.go
Wafffle77: Fixed in 7675031c44179de814c361ca60560b59d2febf7e

pkg/client/smd/smd.go
```diff
@@ -408,6 +409,28 @@ func (sc *SMDClient) GetGroupMembers(group, token string) (client.HTTPEnvelope,
...
 }
 
+// GetGroupMembership is a wrapper function around OchamiClient.GetData that takes
+// a node name, which it passes to the GetData function using the SMD node
+// membership endpoint. It also takes a token, which it puts into the headers as
+// the authorization bearer.
+func (sc *SMDClient) GetGroupMembership(qstr, token string) (client.HTTPEnvelope, error) {
+	if qstr == "" {
+		return client.HTTPEnvelope{}, fmt.Errorf("GetGroupMembership(): node name cannot be empty")
+	}
```
synackd: We'll probably want to clarify this error message since `qstr` can contain other various filters other than the node xname.

cmd/smd/group/membership.go
synackd: Don't forget `groupGetCmd` -> `groupMembershipCmd` ;)

cmd/smd/group/membership.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/OpenCHAMI/ochami/internal/cli"
+	"github.com/OpenCHAMI/ochami/internal/log"
+	"github.com/OpenCHAMI/ochami/pkg/client"
+
+	smd_lib "github.com/OpenCHAMI/ochami/internal/cli/smd"
+)
+
+func newCmdGroupMembership() *cobra.Command {
+	// groupGetCmd represents the "smd group membership" command
+
```
synackd: We can get rid of this extra line to be consistent with the other commands. It is meant to describe the variable below.

cmd/smd/group/membership.go
```diff
@@ -0,0 +1,131 @@
...
+	var groupMembershipCmd = &cobra.Command{
+		Use:   "membership --id <xname> [Other Flags]",
```
synackd: Cobra actually adds `[flags]` to the end of this string so we can get rid of the `[Other Flags]` bit. Currently the usage is printing:

```
Usage:
  ochami smd group membership --id <xname> [Other Flags] [flags]
```

cmd/smd/group/membership.go
```diff
@@ -0,0 +1,131 @@
...
+		Args:  cobra.NoArgs,
+		Short: "Get all group memberships of a node",
+		Long: `Get all group memberships of a node.
```
synackd: Can we replace "of a node" with "of one or more nodes" since >1 nodes are supported now?

cmd/smd/group/membership.go
```diff
@@ -0,0 +1,131 @@
...
+	groupMembershipCmd.Flags().StringArray("subtype", nil, "Filter the results based on HMS subtype. Can be specified multiple times for selecting entries with different subtypes.")
+	groupMembershipCmd.Flags().StringArray("arch", nil, "Filter the results based on architecture. Can be specified multiple times for selecting components with different architectures.")
+	groupMembershipCmd.Flags().StringArray("class", nil, "Filter the results based on HMS hardware class. Can be specified multiple times for selecting entries with different classes.")
+	groupMembershipCmd.Flags().StringArray("nid", nil, "Filter the results based on NID. Can be specified multiple times for selecting entries with multiple specific NIDs.")
+	groupMembershipCmd.Flags().String("enabled", "", "Filter the results based on enabled status (true or false).")
+	groupMembershipCmd.Flags().String("nid_start", "", "Filter the results based on NIDs equal to or greater than the provided integer.")
+	groupMembershipCmd.Flags().String("nid_end", "", "Filter the results based on NIDs less than or equal to the provided integer.")
```
synackd: Let's use hyphens to separate multi-word flags. `nid-start` and `nid-end`

cmd/smd/group/membership.go
```diff
@@ -0,0 +1,131 @@
...
+	// groupGetCmd represents the "smd group membership" command
+
+	var groupMembershipCmd = &cobra.Command{
+		Use:   "membership --id <xname> [Other Flags]",
+		Args:  cobra.NoArgs,
+		Short: "Get all group memberships of a node",
+		Long: `Get all group memberships of a node.
+
+See ochami-smd(1) for more details.`,
+		Example: `  ochami smd group membership x1000c0s0b0n0`,
```
synackd: Could we add a few more examples? I think at minimum we should have:

- no arguments (returns all)
- one or more arguments with no filter flags
- one or more arguments with filter flags

For example:

```go
		Example: `  # Get group membership for all nodes,
  ochami smd group membership

  # Get group membership for a subset of nodes (two)
  ochami smd group membership x1000c0s0b0n0 x1000c0s1b0n0

  # Get group membership for nodes whose IDs are between 1000
  # and 2000 and are of x86 architecture
  ochami smd group membership --nid-start 1000 --nid-end 2000 --arch X86
```

cmd/smd/group/membership.go
```diff
@@ -0,0 +1,131 @@
...
+	groupMembershipCmd.Flags().StringArray("nid", nil, "Filter the results based on NID. Can be specified multiple times for selecting entries with multiple specific NIDs.")
+	groupMembershipCmd.Flags().String("enabled", "", "Filter the results based on enabled status (true or false).")
+	groupMembershipCmd.Flags().String("nid_start", "", "Filter the results based on NIDs equal to or greater than the provided integer.")
+	groupMembershipCmd.Flags().String("nid_end", "", "Filter the results based on NIDs less than or equal to the provided integer.")
+	groupMembershipCmd.Flags().String("partition", "", "Restrict search to the given partition (p#.#). One partition can be combined with at most one group argument which will be treated as a logical AND. NULL will return components in NO partition.")
+	groupMembershipCmd.Flags().String("group", "", "Restrict search to the given group label. One group can be combined with at most one partition argument which will be treated as a logical AND. NULL will return components in NO groups.")
```
synackd: For the help description, other commands do not capitalize the first letter and don't need to be complete sentences. I think I would rather keep these brief and save extended descriptions for the manual page.

man/ochami-smd.1.sc
synackd: Looks like it's still `*membership* <node>`.

man/ochami-smd.1.sc
```diff
@@ -747,6 +747,94 @@ Subcommands for this command are as follows:
...
+
+	*--group* _group_
+		Restrict search to the given group label. One group can be combined with at most one partition argument which will be treated as a logical AND. NULL will return components in NO groups.
+
+	*-F, --format-output* _format_
+		Output response data in specified _format_. Supported values are:
+
+		- _json_ (default)
+		- _json-pretty_
+		- _yaml_
```
synackd: In the other manual pages, I've tried to keep the flags sorted alphabetically (by long option) to keep things predictable to find. Could we do that here?

Also, the convention in the other pages is to keep the character limit per-line to 80, wrapping as needed. Could we enforce that for these flags too?

cmd/smd/group/membership.go
```diff
@@ -0,0 +1,131 @@
...
+				cli.LogHelpError(cmd)
+				os.Exit(1)
+			} else {
+				fmt.Print(string(outBytes))
+			}
+		},
+	}
+
+	// Create flags
+
```
synackd: Let's get rid of this extra newline as well for consistency.

synackd (changes requested):
Thanks for refactoring this to support multiple nodes. I added comments below based on formatting as I haven't tested this yet (will tomorrow).

Some of these comments may seem nitpicky, but they are not meant to be harsh. They are meant to enforce consistency with other commands.

cmd/smd/group/membership.go
Wafffle77: Fixed in a853069bb26af15d2cdfbf4f3d8f47edfc3eb062

man/ochami-smd.1.sc
Wafffle77: Fixed in 4f2a746c2d6c7344a454c9c329389b7d4661bd55 and 0c4b57fce6556417afd6711c84b769a27ee5b6d9

cmd/smd/group/membership.go
Wafffle77: Fixed in c5c27a1e8b3eb1afcf9034372899cffb53173b94

cmd/smd/group/membership.go
Wafffle77: Fixed in 91e15d9db6c991799ce1ba4ce061ad9c619e80c0

cmd/smd/group/membership.go
Wafffle77: Fixed in 1cc29a0362bf72220bd607f0a9fc5f03c3698c2a

cmd/smd/group/membership.go
Wafffle77: Fixed in 1cc29a0362bf72220bd607f0a9fc5f03c3698c2a

cmd/smd/group/membership.go
Wafffle77: Fixed in 96f0b99e0d84830aee271421de856873a8da0ef7

pkg/client/smd/smd.go
Wafffle77: Fixed in 7ddafeb485780592205d57b6142394db4bcde46b

cmd/smd/group/membership.go
Wafffle77: Fixed in 1cc29a0362bf72220bd607f0a9fc5f03c3698c2a

man/ochami-smd.1.sc
synackd: Looks good. Can we move `--format-output` between `--flag` and `--group`?

pkg/client/smd/smd.go
synackd: Testing this, do we need this check? If the query string is empty, the endpoint will return all memberships, which I think should be expected behavior.

cmd/smd/group/membership.go
```diff
@@ -105,23 +113,22 @@ See ochami-smd(1) for more details.`,
...
+	groupMembershipCmd.Flags().StringArray("type", nil, "filter the results based on HMS type")
+	groupMembershipCmd.Flags().StringArray("state", nil, "filter the results based on HMS state")
+	groupMembershipCmd.Flags().StringArray("flag", nil, "filter the results based on HMS flag value")
+	groupMembershipCmd.Flags().StringArray("role", nil, "filter the results based on HMS role")
+	groupMembershipCmd.Flags().StringArray("subrole", nil, "filter the results based on HMS subrole")
+	groupMembershipCmd.Flags().StringArray("softwarestatus", nil, "filter the results based on software status")
+	groupMembershipCmd.Flags().StringArray("subtype", nil, "filter the results based on HMS subtype")
+	groupMembershipCmd.Flags().StringArray("arch", nil, "filter the results based on architecture")
+	groupMembershipCmd.Flags().StringArray("class", nil, "filter the results based on HMS hardware class")
+	groupMembershipCmd.Flags().StringArray("nid", nil, "filter the results based on NID")
```
synackd: Instead of [`StringArray()`](https://pkg.go.dev/github.com/spf13/pflag#FlagSet.StringArray), [`StringSlice()`](https://pkg.go.dev/github.com/spf13/pflag#FlagSet.StringSlice) would be better because it allows both `--flag arg1 --flag arg2` and `--flag arg1,arg2`.

See, e.g.:

https://github.com/OpenCHAMI/ochami/blob/a853069bb26af15d2cdfbf4f3d8f47edfc3eb062/cmd/smd/rfe/get.go#L137-L142

cmd/smd/group/membership.go
```diff
@@ -21,16 +22,23 @@ import (
...
+		Short: "Get all group memberships of one or more nodes",
+		Long: `Get all group memberships of one or more nodes.
 
 See ochami-smd(1) for more details.`,
-		Example: `  ochami smd group membership x1000c0s0b0n0`,
+		Example: `  # Get group membership for all nodes,
+  ochami smd group membership
+
+  # Get group membership for a subset of nodes (two)
+  ochami smd group membership x1000c0s0b0n0 x1000c0s1b0n0
```
synackd: I think I may have misunderstood the intended usage in the new changes. It seems that instead of passing arguments, it is intended usage is to pass flags as filters instead. So this would become:

```
  # Get group membership for a subset of nodes (two)
  ochami smd group membership --id x1000c0s0b0n0,x1000c0s1b0n0
  ochami smd group membership --id x1000c0s0b0n0 --id x1000c0s1b0n0
```

(see `StringSlice()` comment below)

man/ochami-smd.1.sc
```diff
@@ -747,87 +747,93 @@ Subcommands for this command are as follows:
 		flag will *replace* any existing tags, so be sure any existing tags that
 		need to be kept are passed to this flag.
 
-*membership* <node>
+*membership* --id <xname>
```
synackd: This comment is based on my usage comment above.

It seems that these filter flags are optional. In that case, we should remove `--id` from this line as it's optional just like the rest. Instead, to remain consistent with other commands, this line should become:

```
*membership* [-F _format_] [FILTERS]
```

to indicate that `-F` and any of the filter flags below are optional.

cmd/smd/group/membership.go
```diff
@@ -54,14 +62,14 @@ See ochami-smd(1) for more details.`,
 					os.Exit(1)
 				}
 				for _, v := range values {
-					params.Add(flag, v)
+					params.Add(strings.ReplaceAll(flag, "_", "-"), v)
```
synackd: Looking at `strings.ReplaceAll()`, I think the `_` and `-` should be switched (AFAIU we are replacing hyphens with underscores).

cmd/smd/group/membership.go
```diff
@@ -72,7 +80,7 @@ See ochami-smd(1) for more details.`,
 						cli.LogHelpError(cmd)
 						os.Exit(1)
 					}
-					params.Add(flag, value)
+					params.Add(strings.ReplaceAll(flag, "_", "-"), value)
```
synackd: See comment above.

synackd (changes requested):
Initial review after testing.

man/ochami-smd.1.sc
Wafffle77: Do we need the `[-F _format_] [FILTERS]` since it already adds `[flags]` on the end? Between the format and filters, that covers all the flags in the command.

cmd/smd/group/membership.go
Wafffle77: I had originally decided on StringArray to help prevent any parsing weirdness (like if a filter had a comma in it for whatever reason) and because having to specify the filter flag every time you use it makes visualizing the generated query string more intuitive. I can change it if you think StringSlice is better, though.

cmd/smd/group/membership.go
Wafffle77: Fixed in 741bfe06e014c6028dcbfda63ca6f55deee5c8ef

cmd/smd/group/membership.go
Wafffle77: Fixed in 741bfe06e014c6028dcbfda63ca6f55deee5c8ef

man/ochami-smd.1.sc
Wafffle77: In the meantime I've removed the flags entirely with dd0bcfc8ace66aab1c90d362965fac97d742ac89

cmd/smd/group/membership.go
Wafffle77: Fixed in 0092907ab56acf8aa9f83d4fa7c99649fa0a387e for the StringArray version, but I'll update it if it gets swapped to StringSlice.

cmd/smd/group/membership.go
synackd: AFAIK, none of the filters support commas so I think we can go ahead with `StringSlice`. This behavior exists elsewhere so I think users will expect being able to specify multiple flag values via commas.

man/ochami-smd.1.sc
synackd: Oh, this is just for the manual page and not the binary. Adding the `[flags]` string is only something that cobra does in the help/usage output in the binary. In the manual page, there is no automatic generation so we can specify the usage more succinctly.

synackd: If you take a look at the surrounding commands, you can see some examples.

cmd/smd/group/membership.go
synackd: Tested this and it looks to work.

cmd/smd/group/membership.go
synackd: See https://github.com/OpenCHAMI/ochami/pull/92#discussion_r3343989929. I think we can go ahead and update this unless this is a good argument against it.

man/ochami-smd.1.sc
Wafffle77: Ah, my bad. Should be fixed in 418c37e4f84e0d740c9c28bd13485342cf86b12f

cmd/smd/group/membership.go
Wafffle77: Okay, I've changed it over in 5f60ece073669c1452566e85277fb0398c50b7b0

cmd/smd/group/membership.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/OpenCHAMI/ochami/internal/log"
+	"github.com/OpenCHAMI/ochami/pkg/client"
+
+	smd_lib "github.com/OpenCHAMI/ochami/internal/cli/smd"
+)
+
+func newCmdGroupMembership() *cobra.Command {
+	// groupMembershipCmd represents the "smd group membership" command
+	var groupMembershipCmd = &cobra.Command{
+		Use:   "membership --id <xname>",
```
synackd: Let's update this to be just `membership` (since `[flags]` will be added automatically).

man/ochami-smd.1.sc
```diff
@@ -747,6 +747,104 @@ Subcommands for this command are as follows:
 		flag will *replace* any existing tags, so be sure any existing tags that
 		need to be kept are passed to this flag.
 
+*membership* [-F _format_] [FILTERS]
+	Gets the memberships of a node.
```
synackd: "a node" -> "one or more nodes"

man/ochami-smd.1.sc
```diff
@@ -747,6 +747,104 @@ Subcommands for this command are as follows:
 		flag will *replace* any existing tags, so be sure any existing tags that
 		need to be kept are passed to this flag.
 
+*membership* [-F _format_] [FILTERS]
```
synackd, line 750: One thing I noticed is that `membership` comes after `member` in the help output. To match, can we put `membership` after `member` in this man page?

cmd/smd/group/membership.go
```diff
@@ -0,0 +1,139 @@
...
+				"flag",
+				"role",
+				"subrole",
+				"softwarestatus",
+				"subtype",
+				"arch",
+				"class",
+				"nid",
+			} {
+				values, err := cmd.Flags().GetStringArray(flag)
```
synackd: `GetStringArray()` -> `GetStringSlice()`

Getting:

```
$ ./ochami smd group membership -L basic --id x3000c0s0b3n0,x3000c0s0b4n0
ERROR | membership.go:61 > failed to parse flags error="trying to get stringArray value of flag of type stringSlice"
ERROR | cli.go:631 > see 'ochami smd group membership --help' for long command help
```

synackd (changes requested):
Almost there!

Just a couple of formatting things to address as well as a missed function swap.

cmd/smd/group/membership.go
Wafffle77: Fixed in 10ecb38adee2d1ef9f135b04f69296da1dccbc97

man/ochami-smd.1.sc
Wafffle77: Fixed in a2f3d9a1778f022e1b248f5dc0aca238627d4f77

man/ochami-smd.1.sc
synackd: Disregard. Looked over too fast and missed the fact that `group member` is a command group.

cmd/smd/group/membership.go
Wafffle77: Should be fixed in da47a00744576b489409d0bf16017c5711beb806

synackd (approved):
🚀

PR #103, opened 2026-06-19
Add --param flag to bss boot image set to set the name of the kernel parameter the image is passed with

Wafffle77:
## Pull Request Template

### Checklist

- [X] **REUSE Compliance**:  

### Description

Adds a --param/-p flag to the `bss boot image set` command to allow for using a different kernel parameter name when setting the boot image. The flag defaults to "root" so previous behavior is retained. I'll get documentation added once the behavior is confirmed to address #102 properly.

Example:
```
ochami bss boot image set -p rd.live.image <image>
ochami bss boot image set --param rd.live.image <image>
```

Implements #102

### Type of Change

- [ ] Bug fix  
- [X] New feature  
- [ ] Breaking change  
- [ ] Documentation update  

---

### DIFF
--- a/cmd/bss/boot/image/set.go
+++ b/cmd/bss/boot/image/set.go
@@ -194,7 +194,7 @@ See ochami-bss(1) for more details.`,
 	// Create flags
 	bootImageSetCmd.Flags().StringSliceP("xname", "x", []string{}, "one or more xnames whose boot parameters to set")
 	bootImageSetCmd.Flags().StringSliceP("mac", "m", []string{}, "one or more MAC addresses whose boot parameters to set")
-	bootImageSetCmd.Flags().StringP("param", "p", "root", "set the image parameter name")
+	bootImageSetCmd.Flags().StringP("param", "p", "root", "set the kernel parameter name that the image is passed with")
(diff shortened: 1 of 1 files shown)


synackd:
I ran

```
ochami bss boot image set --mac <mac> --param rd.live.image <image>
```

for a `<mac>` that has an existing `root=` parameter and it added the `rd.live.image=` param to the end, but kept the existing `root=` parameter.

I'm thinking we should ensure the `root=` is deleted. This should be straightforward with [`k.DeleteKarg("root")`](https://pkg.go.dev/github.com/synackd/go-kargs#Kargs.DeleteKarg).

synackd:
Being able to modify arbitrary kernel cmdline args would be useful. I've opened #105 to track this. For the scope of this PR, implementing the changes above should be sufficient.

Wafffle77:
Should we make the boot image parameter deletion use the --param flag as well, instead of hard coding "root"? Otherwise I think we run into the same issue you're describing where the parameter isn't properly removed, but with the name of the old image parameter instead. Would it be better to maybe scrap this PR in favor of #105 for just manipulating arbitrary parameters, since that provides the same functionality in a more flexible manner?

PR #108, opened 2026-06-30
feat: add log message color configuration

Wafffle77:
## Pull Request Template

### Checklist

- [X] **REUSE Compliance**:  

### Description

Adds a `--log-color` option for controlling colored log output. In addition to "on" and "off", "auto" (the default) will detect whether or not stderr is  a terminal and act accordingly.

### Type of Change

- [X] Bug fix  
- [X] New feature  
- [ ] Breaking change  
- [ ] Documentation update  

---


synackd (changes requested):
I tested this and it works well.

However, there is a missing piece that is required for config merging to work correctly. For instance, if I:
```
# Set value in user config
ochami config set log.color on
```
I can see it get changed in the user config:
```
$ ochami config show log.color --user
on
```

But the merged config uses the default value:

```
$ ./ochami config show log.color
auto
```

To fix this, we'll have to add a case for this config value in [the custom config unmarshaller](https://github.com/OpenCHAMI/ochami/blob/6f61b17f64e829de6b0436a869ebf8f0260bf958/internal/config/config.go#L88). There is already code there for `timeout` ([case statement](https://github.com/OpenCHAMI/ochami/blob/6f61b17f64e829de6b0436a869ebf8f0260bf958/internal/config/config.go#L105-L124), [default value setting](https://github.com/OpenCHAMI/ochami/blob/6f61b17f64e829de6b0436a869ebf8f0260bf958/internal/config/config.go#L136-L138)) that can serve as a reference.

synackd:
I adjusted the title of this PR to follow Conventional Commit format for the future squash merge.

Wafffle77:
> To fix this, we'll have to add a case for this config value in [the custom config unmarshaller](https://github.com/OpenCHAMI/ochami/blob/6f61b17f64e829de6b0436a869ebf8f0260bf958/internal/config/config.go#L88). There is already code there for `timeout` ([case statement](https://github.com/OpenCHAMI/ochami/blob/6f61b17f64e829de6b0436a869ebf8f0260bf958/internal/config/config.go#L105-L124), [default value setting](https://github.com/OpenCHAMI/ochami/blob/6f61b17f64e829de6b0436a869ebf8f0260bf958/internal/config/config.go#L136-L138)) that can serve as a reference.

Is this desired behavior, though? The other two logging options don't have any default handling in that style, instead the program exits with an error if they aren't specified, and the defaulting appears to be handled by virtue of the options being specified in the default config. I've implemented a quick fix that solves the issue (outside of the config unmarshaling code, since the ConfigLog struct doesn't already have an UnmarshalYAML method).

internal/log/log.go
```diff
@@ -44,6 +45,17 @@ func Init(ll, lf string) error {
...
+
+	if lc == "auto" || lc == "" {
+		cw.NoColor = !term.IsTerminal(int(os.Stderr.Fd()))
+	} else if lc == "on" {
+		cw.NoColor = false
+	} else if lc == "off" {
+		cw.NoColor = true
+	} else {
+		return fmt.Errorf("invalid log-color: %s", lc)
+	}
```
synackd: Could we make this a switch-case?

```go
	switch lc {
	case "", "auto":
		cw.NoColor = !term.IsTerminal(int(os.Stderr.Fd()))
	case "on":
		cw.NoColor = false
	case "off":
		cw.NoColor = true
	default:
		return fmt.Errorf("invalid log-color: %s", lc)
	}

```

synackd (changes requested):

synackd:
> > To fix this, we'll have to add a case for this config value in [the custom config unmarshaller](https://github.com/OpenCHAMI/ochami/blob/6f61b17f64e829de6b0436a869ebf8f0260bf958/internal/config/config.go#L88). There is already code there for `timeout` ([case statement](https://github.com/OpenCHAMI/ochami/blob/6f61b17f64e829de6b0436a869ebf8f0260bf958/internal/config/config.go#L105-L124), [default value setting](https://github.com/OpenCHAMI/ochami/blob/6f61b17f64e829de6b0436a869ebf8f0260bf958/internal/config/config.go#L136-L138)) that can serve as a reference.
> 
> Is this desired behavior, though? The other two logging options don't have any default handling in that style, instead the program exits with an error if they aren't specified, and the defaulting appears to be handled by virtue of the options being specified in the default config. I've implemented a quick fix that solves the issue (outside of the config unmarshaling code, since the ConfigLog struct doesn't already have an UnmarshalYAML method).

The desired behavior is to use the following precedence for configuration options:

1. CLI flag
2. Config from config file passed with `--config`
3. User config file (`~/.config/ochami/config.yaml`)
4. System config file (`/etc/ochami/config.yaml`)
5. Compiled defaults

What I am seeing now is that changing the `log.color` option in (3) does not precede (4) or (5), and the latter two currently take precedence over the former as implemented.

synackd:
Also, would you mind rebasing onto main to resolve the latest changes?

internal/log/log.go
Wafffle77: Fixed in 74d78561009c515cb18b031b327f643fd7890c06

Wafffle77:
> > > To fix this, we'll have to add a case for this config value in [the custom config unmarshaller](https://github.com/OpenCHAMI/ochami/blob/6f61b17f64e829de6b0436a869ebf8f0260bf958/internal/config/config.go#L88). There is already code there for `timeout` ([case statement](https://github.com/OpenCHAMI/ochami/blob/6f61b17f64e829de6b0436a869ebf8f0260bf958/internal/config/config.go#L105-L124), [default value setting](https://github.com/OpenCHAMI/ochami/blob/6f61b17f64e829de6b0436a869ebf8f0260bf958/internal/config/config.go#L136-L138)) that can serve as a reference.
> > 
> > 
> > Is this desired behavior, though? The other two logging options don't have any default handling in that style, instead the program exits with an error if they aren't specified, and the defaulting appears to be handled by virtue of the options being specified in the default config. I've implemented a quick fix that solves the issue (outside of the config unmarshaling code, since the ConfigLog struct doesn't already have an UnmarshalYAML method).
> 
> The desired behavior is to use the following precedence for configuration options:
> 
>     1. CLI flag
> 
>     2. Config from config file passed with `--config`
> 
>     3. User config file (`~/.config/ochami/config.yaml`)
> 
>     4. System config file (`/etc/ochami/config.yaml`)
> 
>     5. Compiled defaults
> 
> 
> What I am seeing now is that changing the `log.color` option in (3) does not precede (4) or (5), and the latter two currently take precedence over the former as implemented.

This looks like a bug in the code that loads all the logging config. I'm able to reproduce the same behavior with the log.format and log.level options as well. A fix might be better relegated to its own PR.

synackd:
> > > > To fix this, we'll have to add a case for this config value in [the custom config unmarshaller](https://github.com/OpenCHAMI/ochami/blob/6f61b17f64e829de6b0436a869ebf8f0260bf958/internal/config/config.go#L88). There is already code there for `timeout` ([case statement](https://github.com/OpenCHAMI/ochami/blob/6f61b17f64e829de6b0436a869ebf8f0260bf958/internal/config/config.go#L105-L124), [default value setting](https://github.com/OpenCHAMI/ochami/blob/6f61b17f64e829de6b0436a869ebf8f0260bf958/internal/config/config.go#L136-L138)) that can serve as a reference.
> > > 
> > > 
> > > Is this desired behavior, though? The other two logging options don't have any default handling in that style, instead the program exits with an error if they aren't specified, and the defaulting appears to be handled by virtue of the options being specified in the default config. I've implemented a quick fix that solves the issue (outside of the config unmarshaling code, since the ConfigLog struct doesn't already have an UnmarshalYAML method).
> > 
> > 
> > The desired behavior is to use the following precedence for configuration options:
> > ```
> > 1. CLI flag
> > 
> > 2. Config from config file passed with `--config`
> > 
> > 3. User config file (`~/.config/ochami/config.yaml`)
> > 
> > 4. System config file (`/etc/ochami/config.yaml`)
> > 
> > 5. Compiled defaults
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What I am seeing now is that changing the `log.color` option in (3) does not precede (4) or (5), and the latter two currently take precedence over the former as implemented.
> 
> This looks like a bug in the code that loads all the logging config. I'm able to reproduce the same behavior with the log.format and log.level options as well. A fix might be better relegated to its own PR.

Hmm... Are you able to file a bug? Being able to reproduce it would be helpful.

synackd:
Created a bug report based on your findings, thanks for finding that!

Will merge this and address the merge issues there.

synackd (approved):

PR #63, opened 2025-11-20
feat: add new discover format and conversion script

synackd:
## Pull Request Template

Thank you for your contribution! Please ensure the following before submitting:

### Checklist

- [x] My code follows the style guidelines of this project  
- [x] I have added/updated comments where needed  
- [x] I have added tests that prove my fix is effective or my feature works  
- [x] I have run `make test` (or equivalent) locally and all tests pass  
- [x] **DCO Sign-off**: All commits are signed off (`git commit -s`) with my real name and email  
- [ ] **REUSE Compliance**:  
  - [ ] Each new/modified source file has SPDX copyright and license headers  
  - [ ] Any non-commentable files include a `<filename>.license` sidecar  
  - [ ] All referenced licenses are present in the `LICENSES/` directory  

#### Feature

Adds a new discovery format along with relevant documentation. Deprecates the old discovery format (but still supports it until it is removed) and provides a conversion script (`extras/scripts/ochami-discovery-old2new.py`) to convert data in the old format into the new one.

The old format consisted of a single `nodes` array, with each node item specifying BMC information (`bmc_mac`, `bmc_ip`, `bmc_fqdn`. However,  with some hardware, there can be multiple nodes per BMC which would mean (1) that the same BMC information must be specified multiple times, once per each node attached to it, and (2) that `ochami` would send duplicate RedfishEndpoint requests (fixed in #62).

The new format consists of the same `nodes` array, but with the `bmc_*` keys removed. Instead, a new `bmcs` array specifies BMC IP, MAC, optional FQDN, xname, and optonal name. The old `bmc_*` options for each node are replaced with a single `bmc` key that references a BMC by its xname or its name, if specified. If `bmc` is not present for a node, `ochami` will attempt to infer the matching BMC based on the node's xname (e.g. node `x1000c0s0b0n2` would be mapped to BMC `x1000c0s0b0`, and so would `x1000c0s0b0n1).

Also, the `group` (string) directive is removed in the new format in favor of `groups` (array).

`ochami` will still detect and use the old discovery format if used, but will warn the user that it is deprecated. The old discovery format is planned to be removed in a future `ochami` release and the old `group` directive is planned to be removed then as well.

Fixes #55

#### Examples

A prime of example of the old format with multiple BMCs is:

```yaml
nodes:
- name: de01
  nid: 1
  xname: x1000c0s0b0n0
  bmc_mac: de:ad:be:ef:00:01
  bmc_ip: 172.16.0.101
[...]

PR #109, opened 2026-06-30
fix: remove duplicate reads in the static discovery subcommand to avoid reading past EOF when using stdin

Wafffle77:
## Pull Request Template

### Checklist

- [X] **REUSE Compliance**:  

### Description

Fixes an issue where some subcommands (tested with `ochami discover static`) would attempt to read a data file multiple times. If the data file was stdin, this would result in the file being read successfully once and subsequent reads would be empty. This commit patches the internal `io.ReadStdin()` to buffer its first successful read from stdin (Full read until EOF, not just a single `Read()` call) and return the buffered data on subsequent calls. This may break anything that expects this behavior from `io.ReadStdin()`, but I didn't see anything that relied on it and all the tests came back fine.

### Type of Change

- [X] Bug fix  
- [ ] New feature  
- [ ] Breaking change  
- [ ] Documentation update  

---

### DIFF
--- a/cmd/discover/static/static.go
+++ b/cmd/discover/static/static.go
@@ -6,6 +6,7 @@
 package static
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
@@ -117,17 +118,32 @@ See ochami-discover(1) for more details.`,
 
 			// Read data from file or stdin into map to determine which
 			// discovery method to use.
-			useDeprecatedFormat := discoverStaticDeprecatedFormat(cmd)
+			discoveryData := make(map[string]([]map[string]any))
+			if cmd.Flag("data").Changed {
+				cli.HandlePayload(cmd, &discoveryData)
+			} else {
+				cli.HandlePayloadStdin(cmd, &discoveryData)
+			}
+			useDeprecatedFormat := discoverStaticDeprecatedFormat(cmd, discoveryData)
+			var rawData []byte
 			if useDeprecatedFormat {
 				log.Logger.Warn().Msg("using deprecated discovery format which will be removed in a future version")
 
-				// Read data from file or stdin
-				nodes := discover.NodeListDeprecated{}
-				if cmd.Flag("data").Changed {
-					cli.HandlePayload(cmd, &nodes)
-				} else {
-					cli.HandlePayloadStdin(cmd, &nodes)
+				// Convert discovery data to struct
+				rawData, err := json.Marshal(discoveryData)
+				if err != nil {
+					log.Logger.Error().Err(err).Msg("unable to marshal discovery data to json")
+					// cli.LogHelpError(cmd)
+					os.Exit(1)
 				}
+				nodes := discover.NodeListDeprecated{}
+				err = json.Unmarshal(rawData, &nodes)
+				if err != nil {
+					log.Logger.Error().Err(err).Msg("unable to unmarshal discovery data from json")
+					// cli.LogHelpError(cmd)
+					os.Exit(1)
+				}
+
 				log.Logger.Debug().Msgf("read %d nodes", len(nodes.Nodes))
 				log.Logger.Debug().Msgf("nodes: %s", nodes)
 
@@ -144,7 +160,6 @@ See ochami-discover(1) for more details.`,
 
 				// Put together payload for different endpoints
 				log.Logger.Debug().Msg("generating redfish structures to send to SMD")
-				var err error
 				comps, rfes, ifaces, err = discover.DiscoveryInfoV2Deprecated(smdBaseURI, nodes)
 				if err != nil {
 					log.Logger.Error().Err(err).Msg("failed to construct structures to send to SMD")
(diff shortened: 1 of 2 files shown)


synackd:
Good catch! I've examined the code and can see the duplicate payload handling.

I'm a bit hesitant to introduce a global variable here and I think the better solution might be to deduplicate the `cli.HandlePayload()`/`cli.HandlePayloadStdin()` calls. For context, these functions appear in both blocks of the `if useDeprecatedFormat {} else {}` block. [[1]](https://github.com/OpenCHAMI/ochami/blob/bacf3d0d287e9027ba643fc868b3fd14254370b3/cmd/discover/static/static.go#L126-L130) [[2]](https://github.com/OpenCHAMI/ochami/blob/bacf3d0d287e9027ba643fc868b3fd14254370b3/cmd/discover/static/static.go#L158-L162) It also appears in the `discoverStaticDeprecatedFormat()` function [[3]](https://github.com/OpenCHAMI/ochami/blob/bacf3d0d287e9027ba643fc868b3fd14254370b3/cmd/discover/static/static.go#L587), which is called before the if block.

What I propose is putting the

```go
// Read data from file or stdin
nodes := discover.NodeListDeprecated{}
if cmd.Flag("data").Changed {
	cli.HandlePayload(cmd, &nodes)
} else {
	cli.HandlePayloadStdin(cmd, &nodes)
}
```

block _before_ the `if useDeprecatedFormat {} else {}` block and remove it everywhere else.

We would have to unmarshal the data into a `map[string]interface{}` or, more robustly, an `interface{}`, perform the deprecation check, then convert the data into the proper data structure (likely through [`format.MarshalData()`](https://github.com/OpenCHAMI/ochami/blob/stdin-buffer/pkg/format/format.go#L62) into [`format.UnmarshalData()`](https://github.com/OpenCHAMI/ochami/blob/stdin-buffer/pkg/format/format.go#L92)) to be used by the appropriate discovery method (deprecated versus modern).

On the other hand, it might be time to remove the deprecated discovery format code, since it was deprecated in v0.6.0 (#63). This would eliminate this issue entirely while slimming the codebase.

Wafffle77:
I think refactoring/removing the deprecated discovery is the best option, but I also wanted to account for any other duplicate reads in the code, since AFAIK there isn't any spec on whether or not duplicate reads should be allowed. It might be a good idea as well to add an error to the ReadStdin function if it gets called more than once per invocation (though that also may require a global variable unless we close stdin when we're done)

Wafffle77:
I've removed the global stdin buffer variable in favor of deduplicating the stdin reads. Long term it might be a good idea to refactor the data/payload system to make it a bit harder for duplicate reads to happen.

cmd/discover/static/static.go
```diff
@@ -117,17 +118,32 @@ See ochami-discover(1) for more details.`,
...
+				cli.HandlePayloadStdin(cmd, &discoveryData)
+			}
+			useDeprecatedFormat := discoverStaticDeprecatedFormat(cmd, discoveryData)
+			var rawData []byte
 			if useDeprecatedFormat {
 				log.Logger.Warn().Msg("using deprecated discovery format which will be removed in a future version")
 
-				// Read data from file or stdin
+				// Convert discovery data to struct
+				rawData, err := json.Marshal(discoveryData)
```
synackd, line 133: Can we use [`format.UnmarshalData()`](https://github.com/OpenCHAMI/ochami/blob/eca5f27f4bf7382f341ebd2993e00fc401755e3b/pkg/format/format.go#L92) and [`format.MarshalData()`](https://github.com/OpenCHAMI/ochami/blob/eca5f27f4bf7382f341ebd2993e00fc401755e3b/pkg/format/format.go#L62) here instead?

Also, can we be more generic with the `discoveryData` type and use an `interface{}`? See below for an example using an `interface{}` with both functions:

https://github.com/OpenCHAMI/ochami/blob/e7536ed90db21c1aef35dc8c270c6b42a3077773/pkg/client/client.go#L360-L365

cmd/discover/static/static.go
```diff
@@ -117,17 +118,32 @@ See ochami-discover(1) for more details.`,
...
+				if err != nil {
+					log.Logger.Error().Err(err).Msg("unable to marshal discovery data to json")
+					// cli.LogHelpError(cmd)
```
synackd: Any reason these are commented out?

synackd (changes requested):
Looks better. Left a comment about some small changes I think would improve this patchset.

synackd:
Updated title to follow Conventional Commit formatting for the future squash merge commit message.

cmd/discover/static/static.go
Wafffle77: I can change it to use format.MarshalData/format.UnmarshalData, but the reason I didn't in the first place is because they just call into the Marshal/Unmarshal functions for the given format, and here the format doesn't matter since it's just used as an intermediary to convert the `map[string]([]map[string]any)` to the correct struct type. Also, the reason I went with `map[string]([]map[string]any)` instead of `interface{}` was because the `discoverStaticDeprecatedFormat` needs to be able to check for the presence of a key in the node definitions. I could change the type, but it would need to be cast internally anyway so I just used the actual type.

cmd/discover/static/static.go
Wafffle77: copy/paste typo :)
I didn't think printing the usage would be particularly helpful if this error happened, so I commented it out when I copied it from above and forgot to remove it. Should be resolved in 2d4b9a627f1b68d3fa6fb029f9803d7679a82490

cmd/discover/static/static.go
synackd: > I can change it to use format.MarshalData/format.UnmarshalData, but the reason I didn't in the first place is because they just call into the Marshal/Unmarshal functions for the given format, and here the format doesn't matter since it's just used as an intermediary to convert the `map[string]([]map[string]any)` to the correct struct type.

I think using the library marshalling functions is functionally fine. My original thought was to use our defined format functions in case anything changed, but this wouldn't be affected by that.

> Also, the reason I went with `map[string]([]map[string]any)` instead of `interface{}` was because the `discoverStaticDeprecatedFormat` needs to be able to check for the presence of a key in the node definitions. I could change the type, but it would need to be cast internally anyway so I just used the actual type.

Thinking on this more, using `map[string]([]map[string]any)` seems like the better approach.

synackd:
I think this is fine. Can you rebase onto `main` and resolve any conflicts?

synackd (approved):

PR #118, opened 2026-07-20
feat: make the verbose option print loaded config values before merging

Wafffle77:
### Checklist

- [X] **REUSE Compliance**:  

### Description

This PR makes the verbose option print the values it loaded from each config file before merging them, for debugging purposes.

Fixes #104

### Type of Change

- [ ] Bug fix  
- [X] New feature  
- [ ] Breaking change  
- [ ] Documentation update  

---

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -561,7 +561,7 @@ func LoadGlobalConfigMerged() error {
 			if err != nil {
 				return fmt.Errorf("failed to serialize config: %w", err)
 			}
-			log.EarlyLogger.BasicLogf("Config data: %s", serializedConfig)
+			log.EarlyLogger.BasicLogf("Config data: \n%s", serializedConfig)
 
 			if err := MergeConfigIntoParser(ko, cfgLoaded.Cfg); err != nil {
 				return fmt.Errorf("failed to merge config: %w", err)
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -561,7 +561,7 @@ func LoadGlobalConfigMerged() error {
 			if err != nil {
 				return fmt.Errorf("failed to serialize config: %w", err)
 			}
-			log.EarlyLogger.BasicLogf("Config data: \n%s", serializedConfig)
+			log.EarlyLogger.BasicLogf("config data:\n%s", serializedConfig)
 
 			if err := MergeConfigIntoParser(ko, cfgLoaded.Cfg); err != nil {
 				return fmt.Errorf("failed to merge config: %w", err)
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -584,6 +584,12 @@ func LoadGlobalConfigMerged() error {
 
 	log.EarlyLogger.BasicLog("config files, if any, have been merged")
 
+	if finalConfig, err := GetConfigString(GlobalConfig, "", "yaml"); err != nil {
+		log.EarlyLogger.BasicLogf("warning: failed to marshal config as YAML: %v", err)
+	} else {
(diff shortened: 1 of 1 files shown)


internal/config/config.go
```diff
@@ -531,6 +531,13 @@ func LoadGlobalConfigMerged() error {
 			}
 		} else {
 			log.EarlyLogger.BasicLogf("merging in config from %s", cfgLoaded.File)
+
+			serializedConfig, err := GetConfigString(cfgLoaded.Cfg, "", "yaml")
+			if err != nil {
+				return fmt.Errorf("failed to serialize config: %w", err)
+			}
+			log.EarlyLogger.BasicLogf("Config data: %s", serializedConfig)
```
synackd: Can we put a newline after `Config data:`? looks a bit awkward to have the first YAML key on the same line as the log message.

synackd (changes requested):
Looks good, just a small change.

Also, would it be useful as a user to see the merged config in early verbose mode?

Wafffle77:
> Also, would it be useful as a user to see the merged config in early verbose mode?

This is already in early verbose mode, unless you meant it would be better for it not to be there. I just did it because it was easiest, since AFAIK the real logger isn't available until the config is parsed.

internal/config/config.go
Wafffle77: Fixed in bc3b1187e0700f1c60a4a7f4e696d91da9f4dd85

synackd: Can we remove the space before the newline?

```go
log.EarlyLogger.BasicLogf("Config data:\n%s", serializedConfig)
```

synackd:
> > Also, would it be useful as a user to see the merged config in early verbose mode?
> 
> This is already in early verbose mode, unless you meant it would be better for it not to be there. I just did it because it was easiest, since AFAIK the real logger isn't available until the config is parsed.

Your change shows the config in each defined config file (system and user), but it doesn't show the final, merged config after everything is loaded.

To do that, we would put after [here](https://github.com/OpenCHAMI/ochami/blob/bc3b1187e0700f1c60a4a7f4e696d91da9f4dd85/internal/config/config.go#L560):

```go
if finalConfig, err := GetConfigString(GlobalConfig, "", "yaml"); err != nil {
	log.EarlyLogger.BasicLogf("warning: failed to marshal config as YAML: %v", err)
} else {
	log.EarlyLogger.BasicLogf("final merged config:\n%s", finalConfig)
}
```

For ease, you can apply this diff (I've tested this):

```diff
diff --git a/internal/config/config.go b/internal/config/config.go
index 7e93335..a5f66dc 100644
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -559,6 +559,12 @@ func LoadGlobalConfigMerged() error {
 
 	log.EarlyLogger.BasicLog("config files, if any, have been merged")
 
+	if finalConfig, err := GetConfigString(GlobalConfig, "", "yaml"); err != nil {
+		log.EarlyLogger.BasicLogf("warning: failed to marshal config as YAML: %v", err)
+	} else {
+		log.EarlyLogger.BasicLogf("final merged config:\n%s", finalConfig)
+	}
+
 	return nil
 }
 
```

internal/config/config.go
synackd: Also just a small nitpick while we are changing this. Can we lower-case the `Config`? Just so it matches the case of the other verbose log messages.

```go
log.EarlyLogger.BasicLogf("config data:\n%s", serializedConfig)
```

synackd: Addressed in 6480817.

synackd (approved):

PR #131, opened 2026-07-29
feat: differentiate simple/advanced API for fabrica payloads

synackd:
### Description

Implement API changes from https://github.com/OpenCHAMI/fabrica/pull/64.

Add `--envelope` flag to fabrica resource commands to allow the user to specify additional metadata that envelops the spec, which contains the actual data. By default (without `--envelope`), the payload is just the spec and additional metadata is ignored.

Fixes #115 

### Checklist

- [x] I have updated the relevant documentation (CLI examples, man pages, README, other docs, etc.)
- [x] **REUSE Compliance**:  
  - [x] Each new/modified source file has SPDX copyright and license headers  
  - [x] Any non-commentable files include a `<filename>.license` sidecar  
  - [x] All referenced licenses are present in the `LICENSES/` directory  

### Type of Change

- [ ] Bug fix  
- [x] New feature  
- [ ] Breaking change  
- [ ] Documentation update  
- [ ] Dependency update

---

### DIFF
--- a/cmd/boot/bmc/add.go
+++ b/cmd/boot/bmc/add.go
@@ -10,6 +10,8 @@ import (
 	boot_service_client "github.com/openchami/boot-service/pkg/client"
 	"github.com/spf13/cobra"
 
+	api "github.com/openchami/boot-service/apis/boot.openchami.io/v1"
+
 	"github.com/OpenCHAMI/ochami/internal/cli"
 	boot_service_lib "github.com/OpenCHAMI/ochami/internal/cli/boot_service"
 	"github.com/OpenCHAMI/ochami/internal/log"
@@ -27,12 +29,17 @@ See ochami-boot(1) for more details.`,
 		Example: `  # Add BMC using payload data
   ochami boot bmc add -d \
     '{
-       "xname": "x1000c0s0b0",
-       "description": "This node's BMC",
-       "interface": {
-         "type": "management",
-         "mac": "de:ca:fc:0f:fe:e1",
-         "ip": "172.16.0.254"
+       "metadata": {
+         "name": "x1000c0s0b0"
+       },
+       "spec": {
+         "xname": "x1000c0s0b0",
+         "description": "This node's BMC",
+         "interface": {
+           "type": "management",
+           "mac": "de:ca:fc:0f:fe:e1",
+           "ip": "172.16.0.254"
+         }
        }
      }'
 
@@ -40,24 +47,48 @@ See ochami-boot(1) for more details.`,
   ochami boot bmc add -d \
     '[
        {
-         "xname": "x1000c0s0b0",
-         "description": "Node 1's BMC",
-         "interface": {
-           "type": "management",
-           "mac": "de:ca:fc:0f:fe:e1",
-           "ip": "172.16.0.1"
+         "metadata": {
+           "name": "x1000c0s0b0"
+         },
+         "spec": {
+           "xname": "x1000c0s0b0",
+           "description": "Node 1's BMC",
+           "interface": {
+             "type": "management",
+             "mac": "de:ca:fc:0f:fe:e1",
+             "ip": "172.16.0.1"
+           }
          }
        },
        {
-         "xname": "x1000c0s0b1",
-         "description": "Node 2's BMC",
-         "interface": {
-           "type": "management",
-           "mac": "de:ca:fc:0f:fe:e2",
-           "ip": "172.16.0.2"
+         "metadata": {
+           "name": "x1000c0s0b1"
+         },
+         "spec": {
+           "xname": "x1000c0s0b1",
+           "description": "Node 2's BMC",
+           "interface": {
+             "type": "management",
+             "mac": "de:ca:fc:0f:fe:e2",
+             "ip": "172.16.0.2"
+           }
          }
        }
      ]'
+
+  # Add BMC preserving labels/annotations (envelope API)
+  ochami boot bmc add -e -d \
+    '{
+       "metadata": {
+         "name": "x1000c0s0b0",
+         "labels": {
+           "env": "prod"
+         }
+       },
+       "spec": {
+         "xname": "x1000c0s0b0"
+       }
+     }'
 
   # Add BMCs using input payload file
   ochami boot bmc add -d @payload.json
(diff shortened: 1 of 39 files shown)


cmd/boot/bmc/add.go
```diff
@@ -27,38 +29,67 @@ See ochami-boot(1) for more details.`,
 		Example: `  # Add BMC using payload data
   ochami boot bmc add -d \
```
davidallendj, line 31: Shouldn't we have `-e` for the full envelope here?

cmd/boot/bmc/add.go
```diff
@@ -27,38 +29,67 @@ See ochami-boot(1) for more details.`,
...
+         "interface": {
+           "type": "management",
+           "mac": "de:ca:fc:0f:fe:e1",
+           "ip": "172.16.0.254"
+         }
        }
      }'
 
   # Add multiple BMCs using payload data
   ochami boot bmc add -d \
```
davidallendj, line 44: Same as line 28 with the `-e` flag unless I'm misunderstanding how the bulk add works.

synackd: The new API requires a "name" to be passed for each item when using `*Simple()` functions.

In the generated CLI (which doesn't support bulk add), this is passed as a positional argument.

In the ochami CLI, it made sense to include this within the payload structure so that bulk items could be added without complicated positional argument handling. The current implementation puts that in `metadata.name`, since there's a field for that already in the metadata block. However, I can see the potential confusion of having to pass metadata without using `-e`, so I'm wondering what the best way of handling this is.

One way is to use a top-level `name` field in the simple API payload and note that in the documentation. Then, for the advanced API, the `metadata.name` field would be used.

I don't see the above to be a problem, but open to thoughts.

davidallendj: I guess that would work. I just expected `-e` for advanced and no flag without out and if the required `metadata.name` isn't there, then it would return an error.

synackd: Yeah, name is required so perhaps adding it as a required, top-level field in the payload the CLI reads makes sense then.

synackd:
I need to refactor this a bit due to some potential confusion arising from the current simple-vs-advanced API data structure (see https://github.com/OpenCHAMI/ochami/pull/131/changes/BASE..ee8d4d1a4457f7c7ee1fe2d1126bcc2fe5bfd303#r3676368580).

synackd:
@davidallendj I've refactored the code so that the simple API only accepts the spec (without having to specify `spec:`) and accepts a top-level `name:` field. Docs, examples, and tests have also been updated.

cmd/boot/bmc/add.go
synackd: This should be addressed in 150c396.

cmd/boot/bmc/add.go
synackd: See refactor: 150c396

davidallendj:
Alright, let me have another look at it.

davidallendj:
This works how I would expect it to and fails with a validation error when trying to use the simple API to add and works when the `-e` flag is included. 

```bash
[rocky@openchami-testing ochami]$ ./ochami metadata defaults add -d@/etc/openchami/data/cloud-init/ci-defaults.yaml
2026-07-30T17:58:54Z ERR add.go:151 > failed to add cluster defaults error="failed to add cluster defaults \"\" ({Description: BaseURL: CloudProvider: Region: AvailabilityZone: ClusterName: ShortName: NidLength:0 PublicKeys:[]}): API error (400): validation failed: base_url is required; cluster_name is required"
2026-07-30T17:58:54Z ERR cli.go:652 > see 'ochami metadata defaults add --help' for long command help
2026-07-30T17:58:54Z WRN add.go:166 > Cluster defaults addition completed with errors
[rocky@openchami-testing ochami]$ ./ochami metadata defaults add -e -d@/etc/openchami/data/cloud-init/ci-defaults.yaml
```

Here is the file contents that are used in the commands above. 

```bash
[rocky@openchami-testing ochami]$ cat /etc/openchami/data/cloud-init/ci-defaults.yaml
{
  "spec": {
    "base_url": "http://172.16.0.254:8081/cloud-init",
    "cluster_name": "demo",
    "nid-length": 2,
    "public-keys": [
      "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIPhN/tcp4FPcZtym/yvDpeGdbCr0mMH0gA7ULBE1m7sk rocky@openchami-testing.novalocal"
      ],
    "short-name": "de"
  }
}
```

I tried the simple version again with removing the `spec` and it works like expected as well.

```bash
[rocky@openchami-testing ochami]$ ./ochami metadata defaults add -d@/etc/openchami/data/cloud-init/ci-defaults.simple.yaml
[rocky@openchami-testing ochami]$ cat /etc/openchami/data/cloud-init/ci-defaults.simple.yaml 
{
    "base_url": "http://172.16.0.254:8081/cloud-init",
    "cluster_name": "demo",
    "nid-length": 2,
    "public-keys": [
      "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIPhN/tcp4FPcZtym/yvDpeGdbCr0mMH0gA7ULBE1m7sk rocky@openchami-testing.novalocal"
      ],
    "short-name": "de"
}
```

I tried adding the simple data format using the advanced API with `-e` and got the validation error as expected.

```bash
[rocky@openchami-testing ochami]$ ./ochami metadata defaults add -e -d@/etc/openchami/data/cloud-init/ci-defaults.simple.yaml
2026-07-30T18:08:21Z ERR add.go:151 > failed to add cluster defaults error="failed to add cluster defaults {Metadata:{Name: UID: Labels:map[] Annotations:map[] CreatedAt:0001-01-01 00:00:00 +0000 UTC UpdatedAt:0001-01-01 00:00:00 +0000 UTC} Spec:{Description: BaseURL: CloudProvider: Region: AvailabilityZone: ClusterName: ShortName: NidLength:0 PublicKeys:[]} Labels:map[] Annotations:map[]}: API error (400): validation failed: base_url is required; cluster_name is required"
2026-07-30T18:08:21Z ERR cli.go:652 > see 'ochami metadata defaults add --help' for long command help
2026-07-30T18:08:21Z WRN add.go:166 > Cluster defaults addition completed with errors
```

davidallendj (approved):
Code changes looks good, works like expected, and all tests passed.

PR #138, opened 2026-08-04
fix(client): inline embedded specs when parsing YAML

synackd:
### Description

Flat YAML payloads only populated the wrapper's name field for fabrica service resource specs (e.g. `BootConfigSpec`) because `yaml.v3` does not automatically inline anonymous embedded structs. As a result, fields such as `kernel`, `initrd`, `params`, and `macs` were silently ignored and empty specs were sent to the service.

Add the `yaml:",inline"` tag to every simple API spec wrapper so YAML fields are decoded into their embedded boot-service and metadata-service specifications.  Do not add an equivalent JSON tag because `encoding/json` already promotes fields from anonymous embedded structs.

Add regression tests covering flat YAML decoding for boot configurations, BMCs, nodes, cluster defaults, groups, instance information, and WireGuard peers. This ensures the CLI's YAML and JSON payload formats retain the same flat simple-API structure.

### Checklist

- [x] I have updated the relevant documentation (CLI examples, man pages, README, other docs, etc.)
- [x] **REUSE Compliance**:  
  - [x] Each new/modified source file has SPDX copyright and license headers  
  - [x] Any non-commentable files include a `<filename>.license` sidecar  
  - [x] All referenced licenses are present in the `LICENSES/` directory  

### Type of Change

- [x] Bug fix  
- [ ] New feature  
- [ ] Breaking change  
- [ ] Documentation update  
- [ ] Dependency update

---

### DIFF
--- a/pkg/client/boot_service/bmc.go
+++ b/pkg/client/boot_service/bmc.go
@@ -22,8 +22,8 @@ import (
 // upstream spec is wrapped with a "name" field so bulk specs can be added with
 // names specified for each without having to provide them as arguments.
 type BMCSpec struct {
-	Name string `json:"name" yaml:"name"` // Mandatory for adding resource
-	api.BMCSpec
+	Name        string `json:"name" yaml:"name"` // Mandatory for adding resource
+	api.BMCSpec `yaml:",inline"`
 }
 
 // AddBMCs is a wrapper that calls the boot-service client's CreateBMC()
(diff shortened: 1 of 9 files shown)


synackd (commented):
Hotfix for issue found when testing https://github.com/OpenCHAMI/release/pull/50 with https://github.com/OpenCHAMI/openchami.org/pull/100. Tests and checks passing.

PR #148, opened 2026-08-11
fix(config): use more of koanf's builtins instead of custom YAML unmarshaling

### LINKED ISSUE #124: [Bug]: Logging configuration precedence not respected when merging config files
## Summary

The logging configuration options (`log.format`, `log.level`, and `log.color`) do not respect the intended configuration precedence when merging multiple configuration files. User-level configuration settings are incorrectly overridden by system defaults and system-level configuration files.

## Issue Origin

This bug was identified in [PR #108 - Comment](https://github.com/OpenCHAMI/ochami/pull/108#issuecomment-5025240013) while adding the `--log-color` configuration option.

## Expected Behavior

The configuration precedence should follow this order (highest to lowest priority):

1. CLI flags
2. Config file passed with `--config` flag
3. User config file (`~/.config/ochami/config.yaml`)
4. System config file (`/etc/ochami/config.yaml`)
5. Compiled defaults

Any logging option explicitly set at a higher priority level should override the same option at a lower priority level.

## Actual Behavior

When loading and merging configuration files, logging options set in the user config file are being overridden by system config file values and compiled defaults. This reverses the intended precedence, making it impossible to override system-level logging settings with user-level settings.

## Root Cause

The issue occurs in `internal/config/config.go` in the `LoadGlobalConfigMerged()` function (lines 497-579). The problem is:

1. The `ConfigLog` struct (lines 159-162) lacks a custom `UnmarshalYAML()` method to intelligently handle partial configuration merges
2. When a user config file specifies only some logging options (e.g., only `log.level`), the `ConfigLog` struct gets unmarshalled with unset fields
3. During the merge operation, these unset fields in the user config override previously-merged values from the system config
4. The behavior differs from other config fields like `timeout`, which has its own `UnmarshalYAML()` method (lines 86-146) that properly handles default value precedence

The `Config` struct's `UnmarshalYAML()` method demonstrates the correct approach by detecting which keys were explicitly set and only applying defaults when a key was not present in the YAML.

### Setup

1. Create `/etc/ochami/config.yaml`:
   ```yaml
   log:
     format: json
     level: debug
     color: "on"
   ```

2. Create `~/.config/ochami/config.yaml`:
   ```yaml
   log:
     level: info
     color: "off"
   ```

3. Run ochami without any config file flag or logging CLI flags

### Expected Result

The merged configuration should be:
- `log.format`: `json` (from system config, not overridden by user)
- `log.level`: `info` (from user config, higher priority than system)
- `log.color`: `off` (from user config, higher priority than system)

### Actual Result

The actual merged configuration is:
- `log.format`: `rfc3339` (compiled default)
- `log.level`: `warning` (compiled default)
- `log.color`: `auto` (compiled default or system value)

The user config file settings are ignored, and the system config settings are also overridden by defaults.

## Impact

This bug affects:
- `log.format` configuration
- `log.level` configuration
- `log.color` configuration (newly added in PR #108)

Users cannot reliably configure logging behavior using config files, making it difficult to maintain consistent logging settings across different environments (development, staging, production).

## Solution Approach

The `ConfigLog` struct needs a custom `UnmarshalYAML()` method similar to the one implemented for the `Config` struct. This method should:

1. Detect which logging configuration keys were explicitly set in the YAML
2. Only apply default values for keys that were not present in the YAML
3. Preserve values that were merged from higher-priority sources before this unmarshalling step

This approach is already proven to work effectively for the `timeout` configuration in the `Config.UnmarshalYAML()` method.

Wafffle77:
### Checklist

- [X] I have updated the relevant documentation (CLI examples, man pages, README, other docs, etc.)
- [X] **REUSE Compliance**:  

### Description
Reworks the config handling to use more of koanf's builtin config merging, loading, and providers to prevent certain bugs with precedence (#124) and other odd configuration behaviors.

Fixes #124

### Type of Change

- [X] Bug fix  
- [ ] New feature  
- [ ] Breaking change  
- [ ] Documentation update  
- [ ] Dependency update


internal/config/config.go
```diff
@@ -499,165 +395,123 @@ func RemoveFromSlice[T any](slice []T, index int) []T {
...
+		log.EarlyLogger.BasicLogf("successfully loaded key-value pairs from config '%s':", path)
+		for _, k := range ko.Keys() {
+			log.EarlyLogger.BasicLogf("\t%s -> %v", k, ko.Get(k))
+		}
 	}
-	log.EarlyLogger.BasicLog("successfully parsed config bytes")
-
-	// Set results as global for later reference/modification
-	GlobalKoanf = k
-	GlobalConfig = cfg
```
synackd, line 628: We never assign the read config file to `GlobalKoanf` or unmarshal into `GlobalConfig`, which `InitLogging()` expects to work:

https://github.com/OpenCHAMI/ochami/blob/68a549d6034fd582a24789be84f677f738fbc513/internal/cli/cli.go#L201-L203

leading to errors like:

```
$ ./ochami --config doc/config.example.yaml config show default-cluster
ochami: failed to initialized logging: failed to Initialize logger: unknown log level: 
ochami: see 'ochami config show --help' for long command help
exit status 1
```

cmd/config/cluster/cluster-delete.go
```diff
@@ -68,31 +69,40 @@ See ochami-config(5) for details on configuration options.`,
...
-			cfg, err := config.ReadConfig(fileToModify)
+			ko, err := config.ReadConfig(fileToModify)
 			if err != nil {
 				log.Logger.Error().Err(err).Msgf("failed to read config from %s", fileToModify)
 				cli.LogHelpError(cmd)
 				os.Exit(1)
 			}
 
+			var clusters []config.ConfigCluster
+			err = ko.Unmarshal("clusters", clusters)
```
synackd: `ko.Unmarshal()` expects a pointer as the second argument, e.g.

```go
err = ko.Unmarshal("clusters", &clusters)
```

I expect `mapstructure` to reject unmarshalling into the slice itself, causing the loop on line 88 to not run, making deletion non-functional.

internal/config/config.go
```diff
@@ -499,165 +395,123 @@ func RemoveFromSlice[T any](slice []T, index int) []T {
...
+					return fmt.Errorf("cluster #%d from config '%s' is missing a name", i, c.name)
+				}
+				if clusterMap[name] == nil {
+					clusterMap[name] = koanf.NewWithConf(kConfig)
+					err = clusterMap[name].Load(confmap.Provider(DefaultClusterConfigMap, "."), nil)
+					if err != nil {
+						return fmt.Errorf("unable to load default cluster config: %w", err)
+					}
+				}
+				err = clusterMap[name].Load(confmap.Provider(cluster["cluster"].(map[string]any), ""), nil)
```
synackd: We shouldn't use an unchecked type assertion here. An empty `cluster` block, e.g.:

```yaml
clusters:
  - name: c1
```

could cause a panic.

The idiomatic thing to do would be to use switch/case on the type, e.g:

```go
switch cls := cluster["cluster"].(type) {
case map[string]any:
  // valid
  err = clusterMap[name].Load(confmap.Provider(cls, ""), nil)
case nil:
  // error: not present
default:
  // error: unknown type
}

File: internal/config/config.go [hunk 4]
```diff
@@ -499,165 +395,123 @@ func RemoveFromSlice[T any](slice []T, index int) []T {
...
 			if err != nil {
-				return fmt.Errorf("failed to serialize config: %w", err)
+				return fmt.Errorf("unable to unmarshal cluster configs from config '%s': %w", c.name, err)
 			}
-			log.EarlyLogger.BasicLogf("config data:\n%s", serializedConfig)
 
-			if err := MergeConfigIntoParser(ko, cfgLoaded.Cfg); err != nil {
-				return fmt.Errorf("failed to merge config: %w", err)
+			for i, cluster := range kClusterSlice {
+				name, ok := cluster["name"].(string)
```
synackd, line 467: The two erring cases I think we should check here are:

1. If name is not a string
2. If name (type string) is empty.

For both of these, I think the best way to do them would be to do the type assertion switch-case as mentioned in my comment below.

internal/config/config.go
```diff
@@ -499,165 +395,123 @@ func RemoveFromSlice[T any](slice []T, index int) []T {
...
+
+	// For merging purposes, maps name to key-value pairs
+	clusterMap := map[string]*koanf.Koanf{}
+	for _, c := range configsToLoad {
+		k2 := koanf.NewWithConf(kConfig)
+		err = k2.Load(c.provider, c.parser)
+		if errors.Is(err, os.ErrNotExist) { // This an error we can ignore
+			log.EarlyLogger.BasicLogf("config '%s' not found, skipping", c.name)
+		} else if err != nil { // If it gets here something has actually gone wrong
+			return fmt.Errorf("unable to load config '%s': %w", c.name, err)
```
synackd, line 452: The docs [say](https://github.com/OpenCHAMI/ochami/blob/68a549d6034fd582a24789be84f677f738fbc513/internal/config/config.go#L393-L394) that config file loading behavior (not just nonexistence) will skip instead of fail. Are we sure we want to change that behavior here?

internal/config/config.go
synackd, line 891: Since we don't support JSON for the config anymore, we can probably use koanf functions for this and get rid of the YAML import in this file.

internal/config/config.go
```diff
@@ -53,97 +53,40 @@ const (
...
+}
+
+var DefaultClusterConfigMap = map[string]any{
+	"enable-auth": true,
 }
 
 var (
-	GlobalConfig   = DefaultConfig // Global config struct
-	GlobalKoanf    *koanf.Koanf    // Koanf instance for gobal config struct
+	GlobalConfig   = Config{}   // Global config struct
```
synackd: `GlobalConfig Config`

cmd/boot/boot.go
```diff
@@ -38,7 +38,7 @@ See ochami-boot(1) for more details.`,
 
 	// Create flags
 	bootCmd.PersistentFlags().String("api-version", "", "version of service API to use in request")
-	bootCmd.PersistentFlags().Duration("timeout", config.DefaultConfig.Timeout, "timeout duration when making requests")
+	bootCmd.PersistentFlags().Duration("timeout", 30*time.Second, "timeout duration when making requests")
```
synackd: I'm thinking it would be better to have this defined as a constant somewhere that we can `time.ParseDuration()` so we don't have to use literals here.

One idea is to define a function (e.g. `GetDefaultTimeout()` that pulls the value from [here](https://github.com/OpenCHAMI/ochami/blob/68a549d6034fd582a24789be84f677f738fbc513/internal/config/config.go#L61) and returns a `time.Duration` from it (thereby being more DRY).

Not strictly necessary, but might be worth considering refactoring

https://github.com/OpenCHAMI/ochami/blob/68a549d6034fd582a24789be84f677f738fbc513/internal/config/config.go#L57-L64

into a function that returns the default config variable.

internal/config/config.go
```diff
@@ -53,97 +53,40 @@ const (
...
+	GlobalKoanf    *koanf.Koanf // Koanf instance for gobal config struct
 	UserConfigFile string
 
 	// Koanf YAML parser provider
 	configParser = kyaml.Parser()
 
 	// Global koanf struct configuration
-	kConfig = koanf.Conf{Delim: ".", StrictMerge: true}
+	kConfig = koanf.Conf{Delim: "."}
```
synackd: As discussed, it seemed the only reason for removing this was so that `timeout` could be passed a number representing the number of nanoseconds. We decided we want to be explicit with the time units (e.g. "30s", "1h") to avoid confusion. In that case, I think we can keep `StrictMerge: true` here.

internal/config/config.go
```diff
@@ -53,97 +53,40 @@ const (
...
 )
 
 // Config represents the structure of a configuration file.
 type Config struct {
```
synackd, line 86: SLight documentation nitpick here, would be good to write a comment explaining that koanf is smart about omitting empty fields so as not to need `omitempty` in the reflect tags. (Just quickly explaining how that works so future maintainers know why they are absent.)

internal/config/config.go
```diff
@@ -462,30 +350,38 @@ func (ccc *ConfigClusterConfig) GetServiceBaseURI(svcName ServiceName) (string,
...
-}
-
 // RemoveFromSlice removes an element from a slice and returns the resulting
 // slice. The element to be removed is identified by its index in the slice.
 func RemoveFromSlice[T any](slice []T, index int) []T {
 	slice[len(slice)-1], slice[index] = slice[index], slice[len(slice)-1]
 	return slice[:len(slice)-1]
 }
 
+func LoadGlobalConfigDefaultOnly() error {
```
synackd, line 379: Docs nitpick: add go function comment

internal/config/config.go
```diff
@@ -462,30 +350,38 @@ func (ccc *ConfigClusterConfig) GetServiceBaseURI(svcName ServiceName) (string,
...
+	log.EarlyLogger.BasicLog("early verbose log messages activated")
+
+	var err error
+	k := koanf.NewWithConf(kConfig)
+	GlobalKoanf = k
```
synackd: More of a semantic thing versus a functional thing, but I think this would be better placed below line 370.

internal/config/config.go
```diff
@@ -499,165 +395,123 @@ func RemoveFromSlice[T any](slice []T, index int) []T {
...
-		}
-
-		// Generate config struct from bytes via parser
-		_, c, err := GenerateConfigFromBytes(cfgBytes)
-		if err != nil {
-			return fmt.Errorf("failed to parse config bytes: %w", err)
-		}
+	var err error
+	k := koanf.NewWithConf(kConfig)
+	GlobalKoanf = k
```
synackd: Another semantic thing, this would be better placed after the loop (once the configs are merged into `k`) below line 468.

internal/config/config.go
```diff
@@ -499,165 +395,123 @@ func RemoveFromSlice[T any](slice []T, index int) []T {
...
+	type configLoader struct {
+		name     string
+		provider koanf.Provider
+		parser   koanf.Parser
+	}
+
+	configsToLoad := []configLoader{
+		{"default", confmap.Provider(DefaultConfigMap, "."), nil},
+		{"system", file.Provider(SystemConfigFile), configParser},
+		// {"system", file.Provider("doc/config.example.yaml"), configParser},
```
synackd: Just adding a note for later to remove.

internal/config/config.go
```diff
@@ -499,165 +395,123 @@ func RemoveFromSlice[T any](slice []T, index int) []T {
...
-			log.EarlyLogger.BasicLogf("starting with default config")
-			if err := MergeConfigIntoParser(ko, cfgLoaded.Cfg); err != nil {
-				return fmt.Errorf("failed to load default config: %w", err)
+	type configLoader struct {
+		name     string
+		provider koanf.Provider
+		parser   koanf.Parser
+	}
+
+	configsToLoad := []configLoader{
```
synackd, line 438: Docs nitpick: I think it would be good to keep the comment outlining the merge order here for clarification, e.g: 

```go
	// Create a parser and merge configs into it:
	//
	//   1. Default config
	//   2. System config
	//   3. User config
	//
```

internal/config/config.go
```diff
@@ -499,165 +395,123 @@ func RemoveFromSlice[T any](slice []T, index int) []T {
...
+		} else { // Good to go
+			log.EarlyLogger.BasicLogf("successfully loaded key-value pairs from config '%s':", c.name)
+			for _, k := range k2.Keys() {
+				log.EarlyLogger.BasicLogf("\t%s -> %v", k, k2.Get(k))
 			}
-		} else {
-			log.EarlyLogger.BasicLogf("merging in config from %s", cfgLoaded.File)
 
-			serializedConfig, err := GetConfigString(cfgLoaded.Cfg, "", "yaml")
+			// Merge clusters separately
```
synackd: Just adding a note here to add documentation on cluster merge behavior (i.e. that cluster config in user config will be merged with that from system config, overwriting existing values)

internal/config/config.go
```diff
@@ -499,165 +395,123 @@ func RemoveFromSlice[T any](slice []T, index int) []T {
...
-	log.EarlyLogger.BasicLogf("parsing config bytes")
-	k, cfg, err := GenerateConfigFromBytes(cfgBytes)
-	if err != nil {
-		return fmt.Errorf("failed to parse config bytes: %w", err)
+	ko := koanf.NewWithConf(kConfig)
+	err := ko.Load(file.Provider(path), configParser)
+	if errors.Is(err, os.ErrNotExist) { // This an error we can ignore
+		log.EarlyLogger.BasicLogf("config '%s' not found, skipping", path)
+	} else if err != nil { // If it gets here something has actually gone wrong
+		return fmt.Errorf("unable to load config '%s': %w", path, err)
```
synackd, line 534: Same comment as https://github.com/OpenCHAMI/ochami/pull/148/changes#r3762243752

internal/config/config.go
```diff
@@ -703,93 +549,84 @@ func ModifyConfig(path, key string, value interface{}) error {
...
+
 	// Make sure that if setting the cluster name, a cluster with that name
 	// doesn't already exist.
 	if key == "name" {
-		for _, cl := range cfg.Clusters {
-			if cl.Name == value.(string) {
-				return fmt.Errorf("cluster with name %q already exists", cl.Name)
+		if err != nil {
+			return fmt.Errorf("unable to unmarshal clusters: %w", err)
+		}
```
synackd: Move this to right below line 560.

internal/config/config.go
```diff
@@ -703,93 +549,84 @@ func ModifyConfig(path, key string, value interface{}) error {
...
 		if key == "name" {
-			// If key was "name", set default-cluster to "name"
-			// instead of cluster specified in arg.
-			cfg.DefaultCluster = value.(string)
+			s, ok := value.(string)
+			if ok {
+				err = ko.Set("default-cluster", s)
+			} else {
+				err = fmt.Errorf("value '%v' is not a string", value)
+			}
```
synackd, line 669: Same checks here as https://github.com/OpenCHAMI/ochami/pull/148/changes#r3762225341

internal/config/config.go
```diff
@@ -804,26 +641,15 @@ func ModifyConfigCluster(path, cluster, key string, dflt bool, value interface{}
...
 		return fmt.Errorf("failed to read %s for deletion: %w", path, err)
 	}
 
-	// Perform deletion
-	ko := koanf.NewWithConf(kConfig)
-	if err := ko.Load(structs.Provider(cfg, "yaml"), nil); err != nil {
-		return fmt.Errorf("failed to load config from %s: %w", path, err)
-	}
-
 	ko.Delete(key)
```
synackd, line 703: Let's add a key existence check here and inform the user if the key does not exist.

internal/config/config.go
synackd: Yeah, we can get rid of 

https://github.com/OpenCHAMI/ochami/blob/68a549d6034fd582a24789be84f677f738fbc513/cmd/config/show.go#L103

as well.

internal/config/config.go
```diff
@@ -1045,25 +849,24 @@ func GetConfigClusterString(cluster ConfigCluster, key, format string) (string,
...
-		return cfg, fmt.Errorf("failed to read config data: %w", err)
+	if err := ko.Load(file.Provider(path), configParser); err != nil {
+		return ko, fmt.Errorf("failed to read config data: %w", err)
 	}
 
-	return cfg, nil
+	return ko, nil
 }
 
 // WriteConfig takes a path and config file format and writes the current viper
```
synackd: Old viper reference 😆 

We should change this to koanf.

internal/log/log.go
```diff
@@ -142,3 +142,13 @@ func (el BasicLogger) BasicLogf(fstr string, arg ...interface{}) {
...
+
+// If early logging is enabled, passes calls through to the underlying io.Writer
+// Otherwise, it's a noop
+func (el BasicLogger) Write(p []byte) (int, error) {
+	if el.EarlyVerbose {
+		return el.out.Write(p)
+	} else {
+		return len(p), nil
+	}
+}
```
synackd: @Wafffle77 noted this is unused, marking for deletion.

cmd/metadata/metadata.go
```diff
@@ -42,7 +42,7 @@ See ochami-metadata(1) for more details.`,
 
 	// Create flags
 	metadataCmd.PersistentFlags().String("api-version", "", "version of service API to use in request")
-	metadataCmd.PersistentFlags().Duration("timeout", config.DefaultConfig.Timeout, "timeout duration when making requests")
+	metadataCmd.PersistentFlags().Duration("timeout", 30*time.Second, "timeout duration when making requests")
```
synackd: See comment [ere](https://github.com/OpenCHAMI/ochami/pull/148/changes#r3768283050).

cmd/config/cluster/cluster-delete.go
```diff
@@ -68,31 +69,40 @@ See ochami-config(5) for details on configuration options.`,
...
+					ko.Delete(fmt.Sprintf("clusters.%d", idx))
 
 					// If cluster was default, remove default-cluster
-					if cfg.DefaultCluster != "" {
-						if cfg.DefaultCluster == clusterName {
-							cfg.DefaultCluster = ""
+					defaultCluster := ko.String("default-cluster")
+					if defaultCluster != "" {
+						if defaultCluster == clusterName {
+							ko.Set("default-cluster", "")
```
synackd: Re-explore why this was set to empty string versus being unset.

cmd/config/cluster/cluster-delete.go
```diff
@@ -68,31 +69,40 @@ See ochami-config(5) for details on configuration options.`,
...
+			}
+
 			// Fetch existing cluster list config
 			clusterName := args[0]
-			for idx, cluster := range cfg.Clusters {
+			for idx, cluster := range clusters {
 				if cluster.Name == clusterName {
-					cfg.Clusters = config.RemoveFromSlice(cfg.Clusters, idx)
+					clusters = config.RemoveFromSlice(clusters, idx)
+					ko.Delete(fmt.Sprintf("clusters.%d", idx))
```
synackd: Note for later: refactor this to not use numbers in path.

internal/config/config.go
```diff
@@ -1045,25 +849,24 @@ func GetConfigClusterString(cluster ConfigCluster, key, format string) (string,
...
-		return cfg, fmt.Errorf("failed to load config file %s: %w", path, err)
+		return ko, fmt.Errorf("failed to load config file %s: %w", path, err)
 	}
 
 	// Unmarshal koanf data into config struct
-	if err := unmarshalKoanfYAML(ko, &cfg); err != nil {
-		return cfg, fmt.Errorf("failed to read config data: %w", err)
+	if err := ko.Load(file.Provider(path), configParser); err != nil {
+		return ko, fmt.Errorf("failed to read config data: %w", err)
 	}
```
synackd, line 944: I noticed we call `ko.Load()` twice here. We should deduplicate to eliminate race conditions.

internal/config/config.go
```diff
@@ -846,34 +672,42 @@ func DeleteConfigCluster(path, cluster, key string) error {
...
-	var clusterToMod *ConfigCluster
-	for cidx, cl := range cfg.Clusters {
-		if cl.Name == cluster {
-			clusterToMod = &(cfg.Clusters[cidx])
-			break
-		}
+	delim := ko.Delim()
+	if strings.Contains(cluster, delim) {
+		return fmt.Errorf("cluster name '%s' contains delimiter character '%s'", cluster, delim)
 	}
```
synackd, line 732: We should apply this check for `AddConfigCluster()` too so we don't have a case where we can add a `cluster.example` name but are unable to remove it.

internal/config/config.go
synackd: Another potential I can see without `StrictMerge` being true is the allowance of `nil` values to be set for values that shouldn't allow it, e.g. `enable-auth` for clusters.

With `StrictMerge` being false, the following could be allowed:

```yaml
clusters:
  - cluster:
      enable-auth:
timeout:
```

and would overwrite default values.

cmd/config/show.go
```diff
@@ -48,32 +49,32 @@ See ochami-config(5) for details on the configuration options.`,
 		Run: func(cmd *cobra.Command, args []string) {
 			// Get the config from the relevant file depending on the flag,
 			// or the merged config if none.
-			var cfg config.Config
+			var ko *koanf.Koanf
 			var err error
 			format := cmd.Flag("format").Value.String()
 			if cmd.Flags().Changed("system") {
-				cfg, err = config.ReadConfig(config.SystemConfigFile)
+				ko, err = config.ReadConfig(config.SystemConfigFile)
```
synackd: I also noticed that [`ReadConfig()`](https://github.com/OpenCHAMI/ochami/blob/68a549d6034fd582a24789be84f677f738fbc513/internal/config/config.go#L852) does not load defaults (nor should it, because its use is for reading only config values that have changed in the read file). Therefore, using it for showing the config will not show default values.

I propose adding a `ReadConfigWithDefaults()` function that starts a koanf object with the defaults and then reads in the config. Then, `config show` and `config cluster show` could use that to show the read config with default values if keys are unset.

synackd (changes requested):
@Wafffle77 and I sat down and reviewed together.

internal/config/config.go
Wafffle77: I think it's probably better to fail, since there are quite a few legitimate errors that can happen during loading, such as parsing, permissions, typos, etc. It will still skip on missing files, though.

internal/config/config.go
Wafffle77: Fixed in 32ace5dd7d131f20b6910bb1dd0710fcebdf1cf3

cmd/config/cluster/cluster-delete.go
Wafffle77: Fixed in 6cdfdb0ba7427c9d79d8b08d699a7246c960a9e6

internal/config/config.go
Wafffle77: Fixed in 183b103130b78fba6e45643ce99415a62e1008f2

internal/config/config.go
Wafffle77: Fixed in 183b103130b78fba6e45643ce99415a62e1008f2

internal/config/config.go
Wafffle77: Fixed in cb813ce2dd3e7945b623465b14e3f51dbb487968

cmd/boot/boot.go
Wafffle77: Fixed in cb813ce2dd3e7945b623465b14e3f51dbb487968

internal/config/config.go
Wafffle77: This ended up causing a bunch of tests to fail since null/nil is a different type, so it was added in 4a5b70884052cb4d42bdde4ea40f5ef4d6c6537b and reverted in bd0481c433869348168d11ff9df064b54f05c6b5

internal/config/config.go
Wafffle77: Fixed in 81c86f8fcee4cc036d6149ba265536b2c12454bf

internal/config/config.go
Wafffle77: Fixed in 98353151d3ceb1a94d30437299b1e4bb80292475

internal/config/config.go
Wafffle77: Fixed in af17f5a850da1301c293fc6706b36612c600131d

internal/config/config.go
Wafffle77: Fixed in af17f5a850da1301c293fc6706b36612c600131d

internal/config/config.go
Wafffle77: Fixed in 730cf731c10db67664dcfa27955ed765842e9481

internal/config/config.go
Wafffle77: Fixed in 7e9da35f6d4e52a38d890dae347acc5b9ba4fdcb

internal/config/config.go
Wafffle77: Fixed in 7e9da35f6d4e52a38d890dae347acc5b9ba4fdcb

internal/config/config.go
Wafffle77: Fixed in 3013a3f5f8e3c9050d1fd73eb8d231a92a77a1ac

internal/config/config.go
Wafffle77: Fixed in 4d1dd42be03b4a32516da0fdebcab2f633062f78

internal/log/log.go
Wafffle77: Fixed in 6b2ef472f557649a1162b12c5b9cf542d4d02647

cmd/metadata/metadata.go
Wafffle77: Fixed in cb813ce2dd3e7945b623465b14e3f51dbb487968

cmd/config/cluster/cluster-delete.go
Wafffle77: Fixed in 6cdfdb0ba7427c9d79d8b08d699a7246c960a9e6

cmd/config/cluster/cluster-delete.go
Wafffle77: Fixed in 6cdfdb0ba7427c9d79d8b08d699a7246c960a9e6

internal/config/config.go
Wafffle77: Fixed in 50546d7ea1cd367d8fc4c5cd993c475000e720b6

internal/config/config.go
Wafffle77: Fixed in 414eb35a013468ffdc33b1b8161143d57019c15b

cmd/config/show.go
Wafffle77: Fixed in bd0481c433869348168d11ff9df064b54f05c6b5

internal/config/config.go
synackd: That makes sense. As a user, can you think of any other cases besides file nonexistence where not exiting on failure would make sense?

If we are changing this to fail here, we should update the documentation to state that that is what happens now. Starting with [this comment](https://github.com/OpenCHAMI/ochami/blob/68a549d6034fd582a24789be84f677f738fbc513/internal/config/config.go#L393-L394) and anywhere else.

internal/config/config.go
synackd: Looks like a check is commented on lines 699-701 above.

cmd/config/show.go
synackd: What's the reason we are setting `StrictMerge` to `false` again in that commit at [`internal/config/config.go:75`](https://github.com/OpenCHAMI/ochami/commit/bd0481c433869348168d11ff9df064b54f05c6b5#diff-54c7c1af5fa8d5db4dc49f0e8e80e93ba2b1183ba4d5c9e2e5729e6deae6a3cd)?

Also looks like the key check for deletion was commented out on lines 799-701.

Finally, `loanf` -> `koanf` on line 921. I also think we can be a bit more specific with the function comment to say that it loads the defaults _before_ loading the config from the passed file.

cmd/config/cluster/cluster-delete.go
synackd: On lines 88-90, it looks like the match condition is inverted. `found` is set to `true` when the name does _not_ match. I think you might have meant to put the `found := true` in an `else` block.

internal/config/config.go
```diff
@@ -17,11 +17,11 @@ import (
 	"time"
 
 	kyaml "github.com/knadh/koanf/parsers/yaml"
+	"github.com/knadh/koanf/providers/confmap"
 	"github.com/knadh/koanf/providers/file"
-	"github.com/knadh/koanf/providers/rawbytes"
 	"github.com/knadh/koanf/providers/structs"
 	"github.com/knadh/koanf/v2"
-	"gopkg.in/yaml.v3"
+	"go.yaml.in/yaml/v3"
```
synackd: Not sure if we talked about it, but was there a reason for changing this path? `gopkg.in/yaml.v3` is pretty standard among Go packages.

internal/config/config.go
```diff
@@ -1045,40 +903,79 @@ func GetConfigClusterString(cluster ConfigCluster, key, format string) (string,
...
+		if !ok {
+			return nil, fmt.Errorf("cluster '%s' is not a map", cluster["name"])
+		}
+		k.Load(confmap.Provider(c, ""), nil)
+		if err != nil {
+			return nil, fmt.Errorf("unable to load cluster config: %w", err)
+		}
+
+		ko.Unmarshal("", &c)
+		kClusterSlice[i]["cluster"] = c
```
synackd: I think this should be:

```go
if err := k.Load(confmap.Provider(c, ""), nil); err != nil {
    return nil, fmt.Errorf("unable to load cluster config: %w", err)
}
kClusterSlice[i]["cluster"] = k.Raw()
```

1. `k.Load()` never assigns its error to be checked.
2. We likely don't want to unmarshal the global koanf config object `ko` into the cluster on line 962.

internal/config/config.go
```diff
@@ -1045,40 +903,79 @@ func GetConfigClusterString(cluster ConfigCluster, key, format string) (string,
...
+	}
+
+	ko.Set("clusters", kClusterSlice)
```
synackd: `ko.Set()` returns an error and we'll probably want to check it.

synackd:
I went ahead and rebased this onto the current `main` to incorporate some new PR checks and fixes to satisfy said checks.

synackd:
We are getting close. Most everything is resolved, just a few comments remain open that require attention. The build failure is because of Go proxy issues and is not a true failure (tested locally).

internal/config/config.go
synackd: Addressed in dafb73c.

internal/config/config.go
synackd: Addressed in c0c4b70.

internal/config/config.go
synackd: Addressed in 243473b.

synackd:
I need these changes to adjust the integration/unit tests on my other branches so I went ahead and addressed these myself. Going to do some basic manual testing and approve once the tests pass.

synackd (approved):

